PR #3, opened 2023-06-09
(DOCSP-30565): Configure drone pipeline for `chat-server` staging

mongodben:
Configuration of Drone pipeline for chat-server staging builds. Will configure prod builds as we get closer to go live. 

This will work for development purposes. 

Automatically pushes the server to staging env when a change is added to `main` branch.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -92,7 +92,7 @@ steps:
       chart_version: 4.12.3
       add_repos: [mongodb=https://10gen.github.io/helm-charts]
       namespace: docs
-      release: chat-server
+      release: ${DRONE_REPO_NAME}-chat-server
       values: image.tag=git-${DRONE_COMMIT_SHA:0:7}-staging,image.repository=795250896452.dkr.ecr.us-east-1.amazonaws.com/docs/${DRONE_REPO_NAME},ingress.enabled=true,ingress.hosts[0]=chat-server.docs.staging.corp.mongodb.com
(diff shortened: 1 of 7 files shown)


.drone.yml
```diff
@@ -1,43 +1,60 @@
...
+# ---
 kind: pipeline
 type: kubernetes
-name: test
+name: test-chat-server
 
 trigger:
   branch:
     - main
+    - DOCSP-30565 # TODO: remove later when ready to merge
```
mongodben, line 12:
```suggestion
```

Delete before merge. Was for testing purposes.

cbush (approved):

chat-server/src/services/database.ts
```diff
@@ -2,37 +2,37 @@ import { Db, MongoClient, ObjectId } from 'mongodb';
...
+export interface AssetDocument {
+  _id: string;
+  data: BinaryData;
+}
+
+interface PageDocument {
+  _id: ObjectId;
+  page_id: string;
+  filename: string;
+  ast: object;
```
cbush, line 19: Why is this needed?

We'll probably end up merging these definitions soon anyway, right?

Btw, TSC should complain about using `object` as such...

mongodben: this was just to get tsc to compile. 

will be deleted once we start populating with our own code.

and re usage of `object`, pls feel free to disallow that when you set up the tsconfig.

PR #5, opened 2023-06-12
Update Q&A Server OpenAPI Spec

nlarew:
Use a resource-oriented approach (e.g. `POST /conversations`, or `GET /conversations/64876105b8480835ec6d6df4`)

### DIFF
--- a/design_docs/openapi.yml
+++ b/design_docs/openapi.yml
@@ -6,166 +6,229 @@ info:
     name: TBD
     url: TBD
 servers:
-  - url: http://TBD.mongodb.com/v1/ # URL TBD
+  - url: http://TBD.mongodb.com/v1
+security:
+  - ApiKeyAuth: []
 paths:
-  /start:
+  /conversations:
     post:
-      operationId: start
-      summary: Start a new conversation with the chatbot
-      description: |
-        This endpoint is used to start a new conversation with the chatbot.
-        The response contains metadata for the conversation.
-      requestBody:
-        content:
-          application/json:
-            schema:
-              type: object
-              properties:
-                user_ip:
-                  type: string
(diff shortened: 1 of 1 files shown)


design_docs/openapi.yml
```diff
@@ -6,166 +6,229 @@ info:
...
     url: TBD
 servers:
-  - url: http://TBD.mongodb.com/v1/ # URL TBD
+  - url: http://TBD.mongodb.com/v1
+security:
+  - ApiKeyAuth: []
 paths:
-  /start:
+  /conversations:
     post:
```
mongodben, line 14: this is now missing the request body

mongodben, line 14: this is now missing the request body. pls add back

mongodben: or do you want to get IP address through different means? HTTP header?

design_docs/openapi.yml
```diff
@@ -6,166 +6,229 @@ info:
...
-  /respond:
+                $ref: '#/components/schemas/Conversation'
+        400:
+          $ref: '#/components/responses/BadRequest'
+        404:
+          $ref: '#/components/responses/NotFound'
+        500:
+          $ref: '#/components/responses/InternalServerError'
+
+  /conversations/{conversationId}:
```
mongodben, line 31: i don't think we need these endpoints for MVP. i'm ok keeping in spec because at some point it'd probably be good to add, but pls mark w a comment saying out of MVP scope

design_docs/openapi.yml
nlarew: We can use express middleware to pull the IP address from every incoming request - no need to manually specify in a request body.

PR #6, opened 2023-06-12
(DOCSP-30618): Create embeddings service

mongodben:
Create OpenAI embeddings service.

As part of this, also set up the following in the project:

- Cleaned up unit test suite so that works on the tests here. 
- Created `integrations/openai.ts` to handle interface with OpenAI API


chat-server/.env.example
```diff
@@ -1,7 +1,5 @@
-# TODO: include our relevant env vars here
-# Connection string to be used for MongoDB to connect to Atlas
-MONGODB_CONNECTION_URI=mongodb://localhost:27017
 # The port number used for the app server
 PORT=3000
-# DB to connect to
-MONGODB_DB_NAME=docs_chat
+OPENAI_ENDPOINT=https://...
+OPENAI_API_KEY=abc123
+OPENAI_EMBEDDING_MODE=text-embedding-ada-002
```
mongodben, line 5: TODO: update with most recent changes

chat-server/.eslintrc.js
```diff
@@ -1,16 +1,18 @@
 module.exports = {
   env: {
     node: true,
+    "jest/globals": true,
```
cbush, line 4: Does this add jest globals to any file, even beyond tests? Can we make an eslintrc file specific to tests? https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy

chat-server/src/integrations/openai.ts
```diff
@@ -0,0 +1,41 @@
+// TODO: Consider refactoring to use new Azure OpenAI client SDK - https://www.npmjs.com/package/@azure/openai
+import axios from "axios";
+import { CreateEmbeddingResponse } from "openai";
+interface CreateEmbeddingParams {
+  input: string;
+  user: string;
+}
+export class OpenAIClient {
```
cbush, line 8: Minor, but, I think, as a rule, it's easier just to stick with CamelCase that treats acronyms and abbreviations as words...
```suggestion
export class OpenAiClient {
```

chat-server/src/integrations/openai.ts
```diff
@@ -0,0 +1,41 @@
...
+    apiKey: string,
+    apiVersion: string
+  ) {
+    this.resourcePath = basePath + deployment;
+    this.apiKey = apiKey;
+    this.apiVersion = apiVersion;
+  }
+
+  embeddings = {
+    create: async ({ input, user }: CreateEmbeddingParams) => {
```
cbush, line 24: I think it would merit declaring an explicit return type here. I would even recommend creating an exported `interface` (or `type`) for like `EmbeddingsService` or whatever this is, then say:
```js
embeddings: EmbeddingsService = { ... }
```

chat-server/src/routes/conversations.ts
```diff
@@ -50,11 +50,22 @@ conversationsRouter.post(
...
       const latestMessage = conversation[conversation.length - 1];
-      const embedding = await embeddings.createEmbedding(latestMessage.content);
-      const chunks = await database.content.findVectorMatches({ embedding });
+      const { status, embeddings: embeddingRes } =
+        await embeddings.createEmbedding({
+          text: latestMessage.content,
+          userIp: ipAddress,
+        });
+      if (status !== 200) {
+        return res.status(status).json({ error: "Embedding error" });
```
cbush, line 64: Can we get any more detail in these errors? Might help later

chat-server/src/routes/conversations.ts
```diff
@@ -50,11 +50,22 @@ conversationsRouter.post(
...
+      }
+      const chunks = await database.content.findVectorMatches({
+        embedding: embeddingRes!,
```
cbush, line 67: The non-null assertion operator (`!`) doesn't really validate at runtime that embeddingRes is not null or undefined. It just tells the compiler that you are _sure_ it is defined. Since we aren't actually sure, you should add a case for it (like the non-200 status case above) or even a proper `assert()` if embeddingRes being null would be a programmer error.

chat-server/src/services/embeddings.ts
```diff
@@ -1,42 +1,95 @@
-import { Configuration, OpenAIApi } from "openai";
+import { OpenAIClient } from "../integrations/openai";
+import { logger } from "./logger";
+import dotenv from "dotenv";
+import { stripIndent } from "common-tags";
+dotenv.config();
```
cbush, line 5: Is this needed? I thought it was one and done in the index file.

chat-server/src/services/embeddings.ts
```diff
@@ -1,42 +1,95 @@
...
+  abstract createEmbedding({
+    text,
+    userIp,
+  }: EmbeddingParams): Promise<EmbeddingResponse>;
 }
 
-class OpenAIEmbeddingProvider extends EmbeddingProvider {
-  private openaiClient: OpenAIApi;
+export class OpenAIEmbeddingProvider extends EmbeddingProvider {
+  private openaiClient: OpenAIClient;
```
cbush, line 39:
```suggestion
  private openAiClient: OpenAIClient;
```

chat-server/src/services/embeddings.ts
```diff
@@ -1,42 +1,95 @@
...
-  async createEmbedding(text: string) {
-    return await this.embeddingProvider.createEmbedding(text);
+  async createEmbedding({ text, userIp }: EmbeddingParams) {
+    return await this.embeddingProvider.createEmbedding({ text, userIp });
   }
 }
 
 // Abstract interface for embedding provider to make it easier to swap out
 // different providers in the future.
 abstract class EmbeddingProvider {
```
cbush, line 31: Why abstract class? Why not just exported interface?

chat-server/src/services/embeddings.ts
```diff
@@ -1,42 +1,95 @@
...
+    try {
+      const res = await this.openaiClient.embeddings.create(embeddingRequest);
+
+      const embeddings = res.data.data[0].embedding;
+      const { status } = res;
+
+      return { embeddings, status };
+    } catch (err: any) {
+      // Catch axios errors which occur if response 4XX or 5XX
+      if (err.response?.status && err.response?.data?.error) {
```
cbush, line 61: Minor: this would be cleaner if you destructured response out of err so you don't need to keep repeating `err.response...` - I would even destructure status and data out of response if response is valid.

chat-server/tests/integrations/openai.test.ts
```diff
@@ -0,0 +1,43 @@
+import { Axios } from "axios";
+import { OpenAIClient } from "../../src/integrations/openai";
+import dotenv from "dotenv";
+dotenv.config();
```
cbush, line 4: We might not want to be using that env file for a test. Is there a way to have dotenv prefer a ".env.test" if one exists?

cbush (commented):
Would like to see comments reviewed before merge - but looks great!

chat-server/.eslintrc.js
mongodben: yea we can scope the jest globals to test files. 

though, i think using `overrides` with in the same eslintrc file would be the most ESLint-idiomatic way to do it - https://eslint.org/docs/latest/use/configure/configuration-files#how-do-overrides-work

chat-server/src/integrations/openai.ts
mongodben: will refactor

chat-server/src/integrations/openai.ts
mongodben: will add

chat-server/.eslintrc.js
cbush: Yah, either way. That's what I was thinking since my other projects tend to just use `.test.ts` instead of an explicit tests subdirectory.

chat-server/src/services/embeddings.ts
cbush: Oh amazing

chat-server/src/integrations/openai.ts
mongodben: actually refactoring Embedding service to be a completely separate entity from LLM service. doing b/c while they're logically linked with the OpenAI service, they're less linked in azure, being separate deployments.

chat-server/src/integrations/openai.ts
mongodben: updated all the `{O|o}pen{AI|ai}` -> `{O|o}penAi`

chat-server/src/routes/conversations.ts
mongodben: ditto https://github.com/mongodb/docs-chatbot/pull/6/files#r1229943342

chat-server/src/services/embeddings.ts
mongodben: b/c i've been writing a lot of Dart and Kotlin recently 😄 

will refactor

chat-server/tests/integrations/openai.test.ts
mongodben: to the best of my understanding, there's not a simple way to do this. 

the easy way would be to have a separate `.env.test` file. for our purposes here, at least for now, all the values would be the same as the `.env` file, so i don't think we need to add.

mongodben: also while investigating this, learned we can replace these lines with: `import "dotenv/config";`

applied that throughout

chat-server/src/services/embeddings.ts
mongodben: also updated `LlmProvider` in `llm.ts` to follow same pattern

chat-server/src/services/embeddings.ts
mongodben: destructured

chat-server/src/integrations/openai.ts
mongodben: did the class implementing interface. i'm partial to classes + constructors plus got them in some other places, so would like to keep (at least for now)

chat-server/src/services/embeddings.ts
```diff
@@ -1,42 +1,96 @@
...
+  async createEmbedding({ text, userIp }: EmbeddingParams) {
+    return await this.embeddingProvider.createEmbedding({ text, userIp });
   }
 }
 
 // Abstract interface for embedding provider to make it easier to swap out
 // different providers in the future.
-abstract class EmbeddingProvider {
-  abstract createEmbedding(text: string): Promise<number[]>;
+interface EmbeddingProvider {
```
cbush, line 29:
```suggestion
export interface EmbeddingProvider {
```

cbush (approved):

PR #8, opened 2023-06-15
Implement ingest store/database

cbush:


mongodben:
@cbush can you include a brief overview in the PR description of what's included in this PR?

ingest/src/main.ts
```diff
@@ -1,5 +1,10 @@
 import * as yargs from "yargs";
 
+import dotenv from "dotenv";
+
+// Configure dotenv early so env variables can be read in imported files
+dotenv.config();
```
mongodben, line 6: nit:
```suggestion
import "dotenv/config";
```

ingest/src/DatabaseConnection.test.ts
```diff
@@ -0,0 +1,132 @@
+import { strict as assert } from "assert";
+import { ObjectId } from "mongodb";
+import { MongoMemoryReplSet } from "mongodb-memory-server";
```
mongodben, line 3: the memory server stuff is nice!

ingest/src/DatabaseConnection.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+  Create a connection to the database.
+ */
+export const makeDatabaseConnection = async ({
+  connectionUri,
+  databaseName,
+}: {
+  connectionUri: string;
+  databaseName: string;
+}): Promise<DatabaseConnection & PageStore & ChunkStore> => {
```
mongodben, line 24: why are these 3 interfaces (`DatabaseConnection`, `PageStore`, and `ChunkStore`) together in the same function? 

could we separate the three, with `PageStore` and `ChunkStore` accepting a `DatabaseConnection` as an argument? it'd provide a greater degree of separation of concern. 

This is the pattern that we use in the Q&A server for the MongoDB interface stuff.

mongodben (approved):
lgtm w 1 structural question and 1 nit

ingest/src/DatabaseConnection.ts
cbush: I don't see how the -Store interfaces could accept a DatabaseConnection as an argument - I don't want to bake any details about how the "store" works (i.e. database vs. s3 bucket vs. ...) into the interface.

That said, this implementation could separate the actual database connection from the store implementations, but I'm trying to strike the balance here between perfect class implementations and gold plating... I think getting the interfaces right is more important, and it will be easy enough to slice out the database connection implementation later if we want to use it.

cbush: Note that most of the actual database connection implementation is just using MongoDB APIs so having it separated in the implementation is pretty weak. Trying something...

cbush: Yeah, it would require me to encode MongoDB APIs (client and db) into the DatabaseConnection interface, which is less great. I'm choosing to stick with this implementation but we can always adapt it later as needed.

PR #11, opened 2023-06-21
(DOCSP-30575): POST /conversations/ route

mongodben:
Add `POST /conversations/` route. refactored the conversations router code for separation of concern and testability.


chat-server/src/routes/conversations/index.ts
```diff
@@ -0,0 +1,64 @@
...
+}: ConversationsRouterParams<OpenAiStreamingResponse, OpenAiAwaitedResponse>) {
+  const conversationsRouter = Router();
+
+  const createConversationRoute = makeCreateConversationRoute({
+    conversations,
+  });
+  /**
+   * Create new conversation.
+   */
+  conversationsRouter.post("/", createConversationRoute);
```
nlarew, line 39:
```suggestion
  /**
   * Create new conversation.
   */
  conversationsRouter.post("/", makeCreateConversationRoute({ conversations }));
```

Can we inline the call here like we do on the other endpoints?

chat-server/src/routes/conversations/rateMessage.ts
```diff
@@ -0,0 +1,69 @@
...
+    conversationId: string;
+    messageId: string;
+  };
+  body: {
+    conversation: Message[];
+    id: string;
+    ip_address: string;
+    message_index: number;
+    rating: boolean;
+  };
```
nlarew, line 23:
```suggestion
  body: {
    rating: boolean;
  };
```

We only need `rating` here

chat-server/src/routes/conversations/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { Conversation } from "../../services/conversations";
+
+export interface MessageResponse {
+  id: string;
+  role: string;
+  content: string;
+  rating?: boolean;
+  timeCreated: number;
```
nlarew, line 8:
```suggestion
  createdAt: number;
```

We're using `createdAt` (and `updatedAt` etc) in the spec

chat-server/src/routes/conversations/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+    id: conversation._id.toString(),
+    messages: conversation.messages.map((message) => ({
+      id: message.id.toString(),
+      role: message.role,
+      content: message.content,
+      rating: message.rating,
+      timeCreated: message.timeCreated.getTime(),
+    })),
+    timeCreated: conversation.timeCreated.getTime(),
+  };
```
nlarew, line 29:
```suggestion
  return {
    id: conversation._id.toString(),
    messages: conversation.messages.map((message) => ({
      id: message.id.toString(),
      role: message.role,
      content: message.content,
      rating: message.rating,
      createdAt: message.createdAt.getTime(),
    })),
    createdAt: conversation.createdAt.getTime(),
  };
```

We're using `createdAt` (and `updatedAt` etc) in the spec

chat-server/src/services/dataStreamer.ts
```diff
@@ -12,12 +14,15 @@ interface AnswerParams {
...
 }
 
-class DataStreamer {
+export interface DataStreamerServiceInterface {
+  answer(params: AnswerParams): Promise<string>;
+}
+export class DataStreamerService {
   // NOTE: for example streaming data, see https://github.com/openai/openai-node/issues/18#issuecomment-1369996933
   async answer({ res, answer, conversation, chunks }: AnswerParams) {
     // TODO: do stuff with the response
```
nlarew, line 23: Streaming is out of scope for this PR right?

chat-server/src/routes/conversations/index.ts
mongodben: yeah, that was a legacy of some debugging. good catch

design_docs/openapi.yml
```diff
@@ -146,50 +146,44 @@ components:
...
             $ref: "#/components/schemas/Message"
-        user_ip:
-          type: string
-          format: ipv4
-          description: The IP address of the user performing the conversation.
-        createdAt:
-          type: integer
-          description: The date the conversation was created.
-        updatedAt:
+        timeCreated:
```
nlarew, line 177: Why change the spec schema?

chat-server/src/services/dataStreamer.ts
mongodben: yeah, this was just for tsc

design_docs/openapi.yml
mongodben: b/c right now we're storing `timeCreated` in DB so i thought it'd be easier to keep API and DB consistent on this.

nlarew: Why are we using `timeCreated` in the db? The spec schemas for those also use `createdAt` and `updatedAt`.

nlarew: Ultimately it doesn't matter that much but we should try to stick to the specs IMO

mongodben: if there was a reason why, i don't remember. will update everything to `(created|updated)At`

chat-server/src/routes/conversations/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  id: string;
+  messages: MessageResponse[];
+  createdAt: number;
+}
+export function convertConversationToResponse(
+  conversation: Conversation
+): ConversationResponse {
+  conversation.messages.shift(); // Remove the system prompt
+  return {
+    id: conversation._id.toString(),
```
nlarew, line 20:
```suggestion
export interface ConversationResponse {
  _id: string;
  messages: MessageResponse[];
  createdAt: number;
}
export function convertConversationToResponse(
  conversation: Conversation
): ConversationResponse {
  conversation.messages.shift(); // Remove the system prompt
  return {
    _id: conversation._id.toString(),
```

Spec has `_id` for the root-level document ID.

nlarew (approved):
LGTM after one more small suggested change

chat-server/src/routes/conversations/utils.ts
mongodben: ok, changing spec back as well

PR #14, opened 2023-06-23
Move reusable pieces to chat-core

cbush:
- Creates a package `chat-core` and moves many things from `chat-server` into it so they can be reused from other places, namely `ingest`.


chat-server/src/aiConstants.ts
```diff
@@ -0,0 +1,53 @@
...
+export const ASSISTANT_PROMPT: OpenAiChatMessage = {
+  role: "assistant",
+  content: "Hello, I'm a MongoDB chatbot. How can I help you today?",
+};
+export const OPENAI_LLM_CONFIG_OPTIONS: GetChatCompletionsOptions = {
+  temperature: 0.3,
+  maxTokens: 500,
+};
+
+interface GenerateUserPromptParams {
```
mongodben, line 30:
```suggestion
export interface GenerateUserPromptParams {
```

chat-server/package.json
```diff
@@ -22,16 +22,16 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@azure/openai": "^1.0.0-beta.2",
+    "chat-core": "file:../chat-core",
```
mongodben, line 25: this is easier than i expected! thought would need some npm link or smthn like that

chat-server/seed-content/index.ts
```diff
@@ -4,9 +4,7 @@ import { Content } from "../src/services/content";
 import { RecursiveCharacterTextSplitter } from "langchain/text_splitter";
 import { Document } from "langchain/dist/document";
 import GPT3Tokenizer from "gpt3-tokenizer";
-import { embeddings } from "../src/services/embeddings";
-import { ObjectId } from "mongodb";
-import { mongodb } from "../src/integrations/mongodb";
+import { ObjectId, embeddings, mongodb } from "chat-core";
```
mongodben, line 7: as we discussed before, we should refactor these to not be exported singletons..but subject of a future PR there

chat-server/src/aiConstants.ts
mongodben, line 1: nit: can you rename `openAiConstants.ts`? these are specific to our OpenAI API implementation, and would presumably be different if we are using another AI service, like Anthropic (or even something in-house)

at some point, we could create abstractions here that'd work for all AI services (perhaps using or inspired by [LangChain](https://js.langchain.com/docs/)), but that's not what we're doing now.

cbush: I don't think openAiConstants.ts will survive long anyway - they aren't supposed to be constants, we should be able to swap them in and out to compare and test etc.

chat-server/seed-content/index.ts
mongodben, line 7: as we discussed before, we should refactor these to not be exported singletons..but subject of a future PR there.

chat-server/src/aiConstants.ts
mongodben: how do you envision them? as env variables?

mongodben (approved):

chat-server/src/aiConstants.ts
cbush: Maybe. I would put all the prompts together somehow and choose which one to pass in based on some configuration.

PR #19, opened 2023-06-27
Move and refactor EmbeddedContent/Store and clean up core services

cbush:
This started as the E2E implementation of the ingest tool, but got too big just merging pieces into the core library and refactoring.

- Moves EmbeddedContent and Page definitions, the "store" interfaces, and database connection (replaces "ContentService" and related classes)
- Refactors and simplifies EmbedService
- Adds handy env var asserter
- Updates chat-server and ingest to use the refactored versions
- A few renames, such as "findVectorMatches" to "findNearestNeighbors"


chat-core/src/EmbeddedContent.ts
```diff
@@ -0,0 +1,90 @@
+import { PersistedPage } from "./Page";
+
+export interface EmbeddedContent {
```
mongodben, line 3: i think we're also adding `score` per https://github.com/mongodb/docs-chatbot/pull/19/files#diff-e4eb4230c061137b41887df64b518d22d061c49648bc27b90621b206dc08faadR115

chat-core/src/OpenAiEmbedFunc.ts
```diff
@@ -0,0 +1,77 @@
...
+import axios from "axios";
+import { EmbedFunc } from "./EmbedFunc";
+import { logger } from "./services/logger";
+import { stripIndent } from "common-tags";
+import { CreateEmbeddingResponse } from "openai";
+
+/**
+  Creates an OpenAI implementation of the embedding function.
+ */
+export const makeOpenAiEmbedFunc = ({
```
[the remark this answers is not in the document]
mongodben: (actually input params might change..but we can deal with that later)

chat-core/src/integrations/openai.test.ts
```diff
@@ -1,49 +1,10 @@
-import {
-  OpenAiEmbeddingsClient,
-  OpenAiChatClient,
-} from "../../src/integrations/openai";
+import { OpenAiChatClient } from "../../src/integrations/openai";
```
mongodben, line 1: why are we just removing embedding stuff from openai integration but not also llm? 

perhaps future work

chat-server/src/app.ts
```diff
@@ -71,7 +37,41 @@ const reqHandler: RequestHandler = (req, _res, next) => {
...
+    MONGODB_DATABASE_NAME,
+    VECTOR_SEARCH_INDEX_NAME,
+    OPENAI_ENDPOINT,
+    OPENAI_API_KEY,
+    OPENAI_EMBEDDING_DEPLOYMENT,
+    OPENAI_EMBEDDING_MODEL_VERSION,
+  } = assertEnvVars(CORE_ENV_VARS);
+
+  // Create instances of services
+  const mongodb = new MongoDB(
```
mongodben, line 54: in future PR we're killing this `MongoDB` as a self defined type to just use the DB

chat-server/seed-content/index.ts
```diff
@@ -22,18 +22,19 @@ const {
...
-  OPENAI_API_KEY!,
-  OPENAI_EMBEDDING_MODEL_VERSION!
-);
-const openAiEmbeddingProvider = new OpenAiEmbeddingProvider(openaiClient);
-const embeddings = new EmbeddingService(openAiEmbeddingProvider);
+} = assertEnvVars(CORE_ENV_VARS);
+
+console.log("MONGODB_CONNECTION_URI", MONGODB_CONNECTION_URI);
+console.log("MONGODB_DATABASE_NAME", MONGODB_DATABASE_NAME);
+
```
mongodben, line 29:
```suggestion
```

chat-server/src/app.ts
```diff
@@ -93,5 +93,10 @@ export const setupApp = async () => {
   });
   app.use(errorHandler);
 
-  return app;
+  return Object.assign(app, {
+    async close() {
+      await mongodb.close();
+      await store.close();
+    },
```
mongodben, line 100: why separate `mongodb` and `store`? aren't using the same DB?

mongodben (approved):
lgtm. few small things throughout. let's merge so we can put those other PRs on top of this

chat-server/src/app.ts
```diff
@@ -71,7 +37,41 @@ const reqHandler: RequestHandler = (req, _res, next) => {
   next();
 };
 
-export const setupApp = async () => {
+export const makeApp = async (): Promise<
```
mongodben, line 40: to make the app more testable, i think it'd make sense to pass mongodb, conversationService, store, dataStreamer, and embed as args to makeApp(). 

then we can pass these things in in the index.ts function which is the only file that's actually involved in running the app.

chat-core/src/integrations/openai.test.ts
cbush: Yeah, as big as this refactor is, the actual purpose is just to allow ingest to use embed. I left as much of the other stuff alone as I could.

chat-server/src/app.ts
cbush: This should be refactored but refactoring ConversationsService etc. was not in scope

chat-core/src/EmbeddedContent.ts
cbush: Done (but not in the actual EmbeddedContent type) --> https://github.com/mongodb/docs-chatbot/pull/19/commits/7ebd62ffd9683b8538e4f32595538a9b7560d817

PR #21, opened 2023-06-29
(DOCSP-30843): IP Address tracking and validation

mongodben:
Track and validate user IP address.


chat-server/src/app.ts
```diff
@@ -42,12 +42,20 @@ export const errorHandler: ErrorRequestHandler = (err, _req, res, _next) => {
...
   const message = `Request for: ${req.url}`;
-  logger.info(createMessage(message, req.body, reqId));
+  logger.info(
+    createMessage({
+      message,
+      requestBody: req.body,
+      requestId: reqId,
+      ipAddress: ip,
+    })
+  );
```
nlarew, line 58:
```suggestion
  const messageObject = createMessage({
     message,
     requestBody: req.body,
     requestId: reqId,
     ipAddress: ip,
   })
  logger.info(JSON.stringify(messageObject));
```

Getting a type error here because the logger expects a string - not sure why exactly since it has an `any` overload. Do we even need this log?

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -103,7 +111,7 @@ export function makeAddMessageToConversationRoute({
       if (!conversationInDb) {
         return sendErrorResponse(res, 404, "Conversation not found");
       }
-      if (conversationInDb.ipAddress !== ipAddress) {
+      if (!areEquivalentIpAddresses(conversationInDb.ipAddress, ip)) {
```
nlarew, line 114: We have essentially this same block in all three routes. Could we abstract to middleware?

chat-server/src/app.ts
mongodben: hmm. not getting any errs. i think you just need to rerun `npm run build` in chat-core

chat-server/src/routes/conversations/addMessageToConversation.ts
mongodben: interesting idea. doing so would require an extra DB round trip per request. unless we were to add the conversation to the request in the middleware, which would seem a little *funky* to me. it'd also require the DB request to get conversation even in scenarios where the req would otherwise fail before DB req necessary. 

i'm open to adding to request in middleware, though think that could make the code a bit more confusing even if reducing a bit of boilerplate.

curious your thoughts here.

nlarew: Reasonable enough to keep it like this for now. Maybe if/when we add more routes we can put in a more shareable abstraction.

nlarew (approved):

PR #23, opened 2023-06-29
(DOCSP-30863) [Q&A] type check requests

nlarew:


mongodben (changes requested):
please include tests for all added functionality.

mongodben (commented):
also, i think it'd make sense to merge #21 first, and rebase these changes on top of that b/c that will also require changes to the zod objects since they'll all now include the `ip` field

PR #24, opened 2023-06-29
(DOCSP-30953 & DOCSP-30945): Error and edge case handling

mongodben:
Add graceful backend handling of the following edge cases:

- When there are no search results for user query: return a static message saying to try again with a different query
  - note that a side effect of this approach is that the chatbot might not answer a follow up question that doesn't have enough context in it. like if the follow up question is "and what about if my app is offline?" and there are no relevant search results, it'll respond with the static message, instead of taking message in the context of the conversation.
    - **this is a tradeoff**. but i think it's worth it b/c it ensure that conversations don't drift away from content in our docs and end up producing lower quality results.
    - post-MVP, i would like to implement functionality to pre-process the conversation with LLM before feeding it to vector search, which will get around this problem. 
- when there are search results for a query but the LLM isn't working, return a static message + vector search results

Added additional error handling:
- return 504 on response timeout (such as if LLM response hanging)
- Return 400 if conversation gets too long (@nlarew should have FE validation for this case) 
- Also added tests for 404 error scenario and generic 500 response.

Also refactored LLM service using the TypeScript interface/make function pattern for better testability (we were planning on doing this later, but including in this PR).

Additionally update the Search results links to be Markdown-formatted links with `?tck=docs_chatbot` to track which clicks to docs come from chatbot.


chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -46,9 +50,9 @@ export interface AddMessageToConversationRouteParams {
   store: EmbeddedContentStore;
   conversations: ConversationsServiceInterface;
   embed: EmbedFunc;
-  llm: LlmProvider<OpenAiStreamingResponse, OpenAiAwaitedResponse>;
+  llm: Llm<OpenAiStreamingResponse, OpenAiAwaitedResponse>;
   dataStreamer: DataStreamerServiceInterface;
-  findNearestNeighborsOptions?: Partial<FindNearestNeighborsOptions>;
+  findNearestNeighborsOptions: FindNearestNeighborsOptions;
```
cbush, line 55: Why this change?

Even if we require findNearestNeighborsOptions, which I'm not sure we should, we probably still want to allow defaults for most of the fields (hence `Partial<...>`)?

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -203,24 +237,56 @@ export async function getContentForText({
...
     return "";
   }
-  const heading = "\n\nArticles Referenced:\n";
+  const heading = hasHeading ? "\n\nFurther Reading:\n" : "";
   const uniqueLinks = Array.from(new Set(chunks.map((chunk) => chunk.url)));
 
   const linksText =
     uniqueLinks.reduce((acc, link) => {
-      const linkListItem = `${link}\n`;
+      const linkListItem = `[${link}](${link}?tck=docs_chatbot)\n`;
```
cbush, line 289: V. nice.

Btw... we might want to start using [URL](https://nodejs.org/api/url.html) more around the code to ensure correct URL building.

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -203,24 +237,56 @@ export async function getContentForText({
...
 }
 export function generateFurtherReading({
   chunks,
+  hasHeading = true,
 }: GenerateFurtherReadingParams) {
   if (chunks.length === 0) {
     return "";
   }
-  const heading = "\n\nArticles Referenced:\n";
+  const heading = hasHeading ? "\n\nFurther Reading:\n" : "";
```
nlarew, line 284:
```suggestion
  const heading = hasHeading ? "\n\nFurther Reading:\n\n" : "";
```

Markdown renders newlines as spaces in this case - we need a double newline to indicate the start of a new paragraph and actually move the first link to the next line.

chat-server/src/services/llm.ts
```diff
@@ -31,91 +31,85 @@ export type OpenAiStreamingResponse = AsyncIterable<
 >;
 export type OpenAiAwaitedResponse = OpenAiChatMessage;
 
-export class OpenAiLlmProvider
-  implements LlmProvider<OpenAiStreamingResponse, OpenAiAwaitedResponse>
-{
-  private openAiChatClient: OpenAiChatClient;
+interface MakeOpenAiLlmParams {
```
cbush, line 34: Minor, but we are using both "Params" and "Args" in the codebase and should probably unify on one.

chat-server/src/services/llm.ts
```diff
@@ -31,91 +31,85 @@ export type OpenAiStreamingResponse = AsyncIterable<
...
-    this.openAiChatClient = openAiChatClient;
-  }
+export function makeOpenAiLlm({
+  apiKey,
+  deployment,
+  baseUrl,
+}: MakeOpenAiLlmParams): Llm<OpenAiStreamingResponse, OpenAiAwaitedResponse> {
+  const openAiChatClient = new OpenAiChatClient(baseUrl, deployment, apiKey);
+  return {
+    // NOTE: for streaming implementation, see // NOTE: for example streaming data, see https://github.com/openai/openai-node/issues/18#issuecomment-1369996933
```
cbush, line 47:
```suggestion
    // NOTE: for example streaming data, see https://github.com/openai/openai-node/issues/18#issuecomment-1369996933
```

chat-server/src/services/llm.ts
```diff
@@ -31,91 +31,85 @@ export type OpenAiStreamingResponse = AsyncIterable<
...
-      options: { ...OPENAI_LLM_CONFIG_OPTIONS, stream: true },
-    });
-    return completionStream;
+  });
+  return [...messages.slice(0, -1), newestMessageForLlm];
+}
+// TODO: consider adding additional validation that messages follow the pattern
+// system, assistant, user, assistant, user, etc.
+// Are there any other things which we should validate here?
+function validateOpenAiConversation(messages: OpenAiChatMessage[]) {
```
cbush, line 94: I'm not sure I understand the purpose of this function. How would a `OpenAiChatMessage[]` get into an invalid state? That should inform whether to throw exceptions or use assertions. If exceptions are thrown due to user error, how would these errors be handled?

chat-server/src/routes/conversations/addMessageToConversation.ts
mongodben: ah, of course! oops will adjust

chat-server/src/app.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      llm,
+      findNearestNeighborsOptions: {
+        // Default options
+        indexName: "default",
+        path: "embedding",
+        k: 3,
+        minScore: 0.9,
+      },
+    });
+  });
```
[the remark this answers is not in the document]
mongodben: no. it is garbage collected

chat-server/src/app.ts
```diff
@@ -1,29 +1,46 @@
...
+      res,
+      status,
+      err.message || "Internal Server Error"
+    );
   } else {
     logger.error(err);
   }
 };
-// Apply to all logs in the app
+// TODO:(DOCSP-31121) Apply to all logs in the app
```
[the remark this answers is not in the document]
mongodben: personally i find it quite useful if the ticket already exists and it will be completed soon. helps me know where to look.

chat-server/src/routes/conversations/addMessageToConversation.ts
mongodben: i don't recall why i did this. probably part of a debugging cycle. my instinct is also that it shouldn't be changed from what it was.

chat-server/src/routes/conversations/addMessageToConversation.ts
mongodben: good point. will look at refactoring here if it's not too hard. though this logic is going to move to the FE soon, so don't want to spend too much time on it.

chat-server/src/services/llm.ts
mongodben: i believe Param is more accurate (see [this page](https://byjus.com/gate/difference-between-argument-and-parameter-in-c-and-c-plus-plus/#:~:text=or%20Actual%20Parameters.-,What%20is%20Parameter%3F,-The%20variables%20that)). where do we use args?

chat-server/src/services/llm.ts
mongodben: dont recall the original intent here (it was originally a method of the class, so there's not any actual new logic)

thinking about it now: exceptions would most likely be thrown due to incorrectly written code elsewhere in the application, as opposed to user error. 

think it's worth keeping this here to make clear at this module's level what the shape of the conversation must be.

chat-server/src/app.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      databaseName: MONGODB_DATABASE_NAME,
+    });
+    app = await makeApp({
+      embed,
+      store,
+      conversations,
+      dataStreamer,
+      llm,
+      findNearestNeighborsOptions: {
+        // Default options
```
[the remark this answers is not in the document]
mongodben:
```suggestion
```

chat-server/src/routes/conversations/addMessageToConversation.ts
mongodben: fixed. had to do change in a few other places too.

chat-server/src/routes/conversations/addMessageToConversation.ts
mongodben: refactored to use this.

PR #25, opened 2023-06-29
Streaming works

nlarew:


mongodben (changes requested):
please add tests for everything

chat-ui/src/services/conversations.ts
```diff
@@ -61,9 +94,129 @@ export default class ConversationService {
...
+          response.headers.get("content-type") === "text/event-stream"
+        ) {
+          return; // everything's good
+        }
+
+        if (
+          response.status >= 400 &&
+          response.status < 500 &&
+          response.status !== 429
+        ) {
```
mongodben, line 187: what happens if 5XX error?

chat-server/src/services/dataStreamer_class.ts
mongodben, line 1: i think content fo this file isn't used. can we delete?

mongodben (changes requested):
looks great, excellent work nick! 

some small things and questions throughout.

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -193,30 +192,53 @@ export function makeAddMessageToConversationRoute({
...
+        dataStreamer.connect(res);
+        const answer = await dataStreamer.stream({
+          stream: answerStream,
+        });
+        logger.info(`LLM response: ${JSON.stringify(answer)}`);
+        await dataStreamer.streamData({
+          type: "delta",
+          data: furtherReading,
+        });
+        answerContent = answer + furtherReading;
```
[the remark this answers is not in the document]
nlarew: Yeah we have total control over the stream payload so depending on what exactly we want we could either stream `furtherReading` as a markdown delta or add a new `type` like you suggest

PR #27, opened 2023-06-30
Fix tests

cbush:


seed-content/src/transformSample.ts
```diff
@@ -0,0 +1,57 @@
+import { promises as fs } from "fs";
```
mongodben, line 1: can you add a file description at the top here explaining what this script does and context?

nothing formal. i think we'll only need to keep this in repo til staging is updated to the new data model. but would like a little record of what this is.

mongodben (approved):
lgtm

PR #29, opened 2023-07-06
(DOCSP-30890) [UI] Make component mobile responsive

nlarew:
Jira: (DOCSP-30890) [UI] Make component mobile responsive

## Proposed Changes

- Responsive styles for Small, Medium, and Large+ screens ([as defined in Snooty](https://github.com/mongodb/snooty/blob/master/src/theme/docsTheme.js#L39-L47))
- Adds a modal close button

### DIFF
--- a/chat-ui/src/App.module.css
+++ b/chat-ui/src/App.module.css
@@ -1,0 +1,14 @@
+body {
+  margin: 0;
+}
+
+.app_background {
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 20px;
+  background: #001e2b;
+  padding: 20px;
+  height: 100vh;
+}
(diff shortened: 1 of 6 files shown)


mongodben:
there's some x-overflow, though it is scrollable. why not just make the content wrap? 
Screenshot 2023-07-07 at 10 20 28 AM

mongodben (changes requested):
there's some x-overflow, though it is scrollable. can you make the content fit the width and wrap? 
Screenshot 2023-07-07 at 10 20 28 AM

mongodben:
also, on further thought, if you think this work will be rendered irrelevant when we use the leafy green components, i don't think we need to belabor it

nlarew:
@mongodben fixed this for now with a small CSS hack. Should go away / be "not our problem" with the LG Chat components.

PR #34, opened 2023-07-10
Add minor optimization for rate-limited embed requests

cbush:
Jira: N/A

- When hitting the rate limit, use the recommendation in the message (if available) to determine backoff time.
- Run embeddings sequentially, as we're most likely to hit rate limit before any other performance bottleneck. This prevents overload of rate-limited requests and log spam.
- Increase attempt count to 25.

### DIFF
--- a/chat-core/src/OpenAiEmbedFunc.ts
+++ b/chat-core/src/OpenAiEmbedFunc.ts
@@ -99,8 +99,21 @@ export const makeOpenAiEmbedFunc = ({
           }
 
           logger.info(
-            `OpenAI Embedding API rate limited (attempt ${attemptNumber - 1})`
+            `OpenAI Embedding API rate limited (attempt ${
+              attemptNumber - 1
+            }): ${error.message}`
           );
+
+          // Quick optimization for retry where we wait as long as it tells us
+          // to (if it does)
+          const matches = /retry after ([0-9]+) seconds/.exec(error.message);
+          const waitSeconds = matches ? parseInt(matches[1]) : 0;
+          if (waitSeconds) {
+            await new Promise((resolve) =>
+              setTimeout(resolve, waitSeconds * 1000)
+            );
+          }
+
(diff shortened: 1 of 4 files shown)


chat-core/src/OpenAiEmbedFunc.ts
```diff
@@ -90,17 +90,31 @@ export const makeOpenAiEmbedFunc = ({
...
-            `OpenAI Embedding API rate limited (attempt ${attemptNumber - 1})`
+            `OpenAI Embedding API rate limited (attempt ${
+              attemptNumber - 1
+            }): ${errorMessage}`
           );
+
+          // Quick optimization for retry where we wait as long as it tells us
+          // to (if it does)
+          const matches = /retry after ([0-9]+) seconds/.exec(errorMessage);
+          const waitSeconds = matches ? parseInt(matches[1]) : 0;
```
mongodben, line 111: clever!

ingest/src/updateEmbeddedContent.ts
```diff
@@ -76,19 +76,20 @@ export const updateEmbeddedContentForPage = async ({
...
-      };
-    })
-  );
+  const embeddedContent: EmbeddedContent[] = [];
+  // Process sequentially because we're likely to hit rate limits before any
+  // other performance bottleneck
+  for (const chunk of contentChunks) {
+    const { embedding } = await embed({
+      text: chunk.text,
+      userIp: "127.0.0.1",
```
mongodben, line 85: not important for this PR, but can you just get the IP of the system where the script is running?

mongodben (approved):

ingest/src/updateEmbeddedContent.ts
cbush: Possibly, though it would add a dependency on an external site? https://stackoverflow.com/questions/20273128/node-js-how-to-get-my-external-ip-address-in-node-js-app

PR #36, opened 2023-07-12
Fix URL building for devcenter paths

cbush:
Jira: N/A

Fixes the URL builder, which was copied from the Snooty side, but devcenter `calculated_slug` has a leading slash which, it turns out, overrides any path in the baseUrl.


mongodben (approved):
this makes sense to me. 

is there some test that we could add/or update to validate this behaves as expected and defend against any issues like the one that occurred going forward?

PR #37, opened 2023-07-12
Fix + improve devcenter source tests

cbush:
Jira: N/A

- Fixes test given so many discarded results
- Checks URL sanity


mongodben (approved):
nice, ty

PR #40, opened 2023-07-12
(DOCSP-31304): Implement 'all' command

cbush:
Jira: https://jira.mongodb.org/browse/DOCSP-31304

## Changes

- Creates 'all' command that can be run from docker, etc.
- Moves actual command logic to `doCommand` external functions so I can call them from elsewhere
- Uses a meta collection to keep track of previous successful runs - deducing this from existing documents turns out to be pretty complicated and we may as well just use a very early 'since' at that point: we would look at all pages and embeddings to make sure everything is successfully updated, then update those that have not yet been updated (the embed command does resume where it left off at least, so it wouldn't just duplicate work). That said, having a meta document in the database is a pretty simple optimization.
- Bug fix: warn and gracefully handle empty chunks set (which can happen when sources return empty pages)

## Test

- Ran once locally, aborted
- Ran again, started from dawn of time (2023) due to no prior successful run
- Ran again after successful run and it correctly determined it had nothing to update

### DIFF
--- a/ingest/src/commands/all.test.ts
+++ b/ingest/src/commands/all.test.ts
@@ -1,7 +1,8 @@
 import { PageStore, EmbeddedContentStore, assertEnvVars } from "chat-core";
-import { INGEST_ENV_VARS } from "../IngestEnvVars";
-import { doAllCommand, makeIngestMetaStore } from "./all";
 import { MongoClient } from "mongodb";
+import { INGEST_ENV_VARS } from "../IngestEnvVars";
+import { doAllCommand } from "./all";
+import { makeIngestMetaStore } from "../IngestMetaStore";
 
 import "dotenv/config";
 
--- a/ingest/src/commands/all.test.ts
+++ b/ingest/src/commands/all.test.ts
@@ -1,0 +1,115 @@
+import { PageStore, EmbeddedContentStore, assertEnvVars } from "chat-core";
+import { INGEST_ENV_VARS } from "../IngestEnvVars";
+import { doAllCommand, makeIngestMetaStore } from "./all";
+import { MongoClient } from "mongodb";
+
+import "dotenv/config";
+
+jest.setTimeout(1000000);
+
+describe("allCommand", () => {
+  const { MONGODB_CONNECTION_URI: connectionUri } =
+    assertEnvVars(INGEST_ENV_VARS);
+
+  const mockEmbeddedContentStore: EmbeddedContentStore = {
+    async deleteEmbeddedContent() {
+      return;
+    },
+    async findNearestNeighbors() {
+      return [];
+    },
+    async loadEmbeddedContent() {
+      return [];
+    },
+    async updateEmbeddedContent() {
+      return;
+    },
+  };
+  const mockPageStore: PageStore = {
+    async loadPages() {
+      return [];
+    },
+    async updatePages() {
+      return;
+    },
+  };
+
+  let databaseName: string;
+
+  beforeEach(async () => {
+    databaseName = `test-all-command-${Date.now()}-${Math.floor(
+      Math.random() * 10000000
+    )}`;
+  });
+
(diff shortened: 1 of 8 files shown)


ingest/src/commands/all.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      return;
+    },
+  };
+
+  let databaseName: string;
+
+  beforeEach(async () => {
+    databaseName = `test-all-command-${Date.now()}-${Math.floor(
+      Math.random() * 10000000
+    )}`;
```
mongodben, line 42: nit: seems redundant to have both Date.now() and Math.floor() calls. both of these would individually provide the randomness necessary for a unique DB name

ingest/src/commands/embed.ts
```diff
@@ -3,13 +3,15 @@ import {
...
+  EmbeddedContentStore,
+  PageStore,
 } from "chat-core";
 import { updateEmbeddedContent } from "../updateEmbeddedContent";
 import { INGEST_ENV_VARS } from "../IngestEnvVars";
 
 type EmbeddedContentCommandArgs = {
   since: string;
-  source?: string;
+  source?: string | string[];
```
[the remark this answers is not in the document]
mongodben: in the `pages` command, you provide some description https://github.com/mongodb/docs-chatbot/pull/40/files#diff-093bea1d5bc410424c2ca3791901a696e9842f1a179b48e981edac670e3cb57dR26. could you add that here as well?

ingest/src/commands/embed.ts
```diff
@@ -18,45 +20,20 @@ const commandModule: CommandModule<unknown, EmbeddedContentCommandArgs> = {
     return args.string("since").string("source").demandOption("since");
   },
   async handler({ since, source }) {
```
mongodben, line 22: can you do some light data validation on the `since` param? 

something like:

```ts
if(!Date.parse(since)){
  throw new Error(`The value for 'since' (${since}) must be a valid JavaScript date string.`)
}

```

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

ingest/src/commands/all.test.ts
cbush: Because this is a unit test if both run within the same ms (not likely but theoretically possible) we would have a conflict. Better safe than mildly confused and annoyed by a flaky test.

ingest/src/commands/pages.ts
```diff
@@ -16,68 +21,80 @@ const commandModule: CommandModule<
...
+    )
+  );
+
+  const devCenterSource = await makeDevCenterDataSource({
+    name: "devcenter",
+    collectionName: "search_content_prod",
+    databaseName: "devcenter",
+    connectionUri: DEVCENTER_CONNECTION_URI,
+    baseUrl: "https://www.mongodb.com/developer",
+  });
```
mongodben, line 75: thought for future work: would be nice to have these pluggable in some way so they're not defined inside of this function..obvi no action here now

cbush: That is the eventual plan

mongodben (approved):
lgtm w a few small things

ingest/src/commands/all.ts
```diff
@@ -1,12 +1,170 @@
...
+  If the 'since' date given to the embed command is too late, pages that were
+  updated during a failed run will not be picked up.
+
+  If too early, more pages and embeddings will be checked than necessary. The
+  embed command will not unnecessarily create new embeddings for page updates
+  that it has already created embeddings for, but it would still be wasteful to
+  have to check potentially all pages and embeddings when the date is early
+  enough. 
+ */
+export type IngestMetaStore = {
```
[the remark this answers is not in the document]
cbush: Probably not, but it doesn't hurt to be in its own file.

PR #182, opened 2023-09-25
[DOCSP-32954] Add ErrorBanner, React-Transition-Group

jackyw2018:
### Resources

- [Jira Ticket](https://jira.mongodb.org/browse/DOCSP-32954)


chat-ui/src/DevCenterChatbot.tsx
```diff
@@ -361,20 +367,35 @@ function ChatbotModal({
...
+    ...LGInputBarProps
+  } = props;
   const { darkMode } = useDarkMode();
 
+  if (conversationError)
+    return (
+      <div className={styles.chatbot_input_area}>
+        <ErrorBanner darkMode={darkMode} message={conversationError} />
+      </div>
+    );
```
nlarew, line 388: I don't think conversation errors are the concern of the input bar, especially given that we use it here to short circuit rendering. In the docs `Chatbot` component we conditionally render the InputBar based on conversation error state rather than have the InputBar handle the conditional internally.

Could we refactor here to use that pattern instead? It'll make it easier to combine this all into shared components.

chat-ui/src/DevCenterChatbot.tsx
```diff
@@ -526,26 +564,70 @@ const Message = ({
...
+    opacity: 0,
+  };
+
+  const transitionStyles = {
+    entering: { opacity: 1 },
+    entered: { opacity: 1 },
+    exiting: { opacity: 0 },
+    exited: { opacity: 0 },
+    unmounted: { opacity: 0 },
+  };
```
nlarew, line 602: Let's use emotion for this instead of the jsx `style` prop. Also at this point we're not actually using any of these states so RTG isn't adding anything until we wire it up.

Since we're only doing CSS transitions right now I think it'd be easier to switch to [`CSSTransition`](https://reactcommunity.org/react-transition-group/css-transition) - then we don't have to manually process the state. We just right the CSS we want for each state and it's wired up automatically.

You should be able to put all this in the `message_prompts` css (note the different state names compared to regular `Transition`):

```ts
  message_prompts: css`
    margin-left: 70px;
    @media screen and (max-width: 804px) {
      margin-left: 50px;
    }

    transition: opacity ${duration}ms ease-in;

    &-enter {
      opacity: 0;
    }
    &-enter-active {
      opacity: 1;
    }
    &-exit {
      opacity: 1;
    }
    &-exit-active {
      opacity: 0;
    }
  `,
```

then you'll replace `Transition` with `CSSTransition`, add the prop `classNames={styles.message_prompts}`, and then add `ref={nodeRef}` to the div that has the `message_prompts` style.

chat-ui/src/DevCenterChatbot.tsx
```diff
@@ -526,26 +564,70 @@ const Message = ({
...
+
+  return (
+    <Transition in={inProp} timeout={duration} nodeRef={nodeRef}>
+      {(state) => (
+        <div
+          className={styles.message_prompts}
+          style={{ ...defaultStyle, ...transitionStyles[state] }}
+        >
```
nlarew, line 610: Per my earlier comment about using emotion & CSSTransition. Also make sure to import `CSSTransition` too!

```suggestion
    <CSSTransition
      in={inProp}
      timeout={duration}
      nodeRef={nodeRef}
      classNames={styles.message_prompts}
    >
      <div
        className={styles.message_prompts}
        ref={nodeRef}
      >
```

nlarew (changes requested):

PR #187, opened 2023-09-26
Add scrubbed message analyzer

cbush:
Jira: none

## Changes

- Uses TypeChat to analyze original messages and save the analysis to the scrubbed_messages collection before the original messages disappear 
- Refactors our TypeChat usage to move it into a more general use case chat-core

## Notes

- This tries its best to scrub PII, but there's a possibility that PII will still come through. I'm thinking that when we look at the analysis in aggregate, some of the topics or sentiments that might indicate PII will stand out and we can simply delete those entries. I also added a PII flag and in my testing ChatGPT is pretty good at detecting whether the message contains PII in the first place.


scripts/src/analyzeMessages.ts
```diff
@@ -0,0 +1,133 @@
...
+  const conversationsCollection = db.collection<Conversation>("conversations");
+  const originalMessages = conversationsCollection.aggregate<Message>([
+    {
+      // Find messages in a recent timeframe
+      $match: {
+        "messages.createdAt": {
+          // Originals for older conversations are likely to have been deleted,
+          // so we can't use it to get the topic
+          $gte: daysBeforeDate(8),
+        },
```
nlarew, line 48:
```suggestion
        },
        "messages": {
          "$elemMatch": { role: "user" }
        }
```

We can add this so that our initial match only includes conversations that actually have a user message (~30% do not have a user message)

scripts/src/analyzeMessages.ts
```diff
@@ -0,0 +1,133 @@
...
+        localField: "id",
+        foreignField: "_id",
+        as: "scrubbed",
+      },
+    },
+    {
+      // Filter out messages that haven't been scrubbed yet or that already have
+      // analysis set
+      $match: {
+        scrubbed: { $ne: [] },
```
nlarew, line 79:
```suggestion
        "scrubbed.0": { $exists: true },
```

Don't think it's really functionally different but the array index pattern is also an efficient way to check a minimum array size.

scripts/src/scrubMessages.ts
```diff
@@ -73,7 +52,7 @@ const scrubMessages = async ({ db }: { db: Db }) => {
     },
     {
       $project: {
-        _id: "$messages.id",
+        _id: "$messages.id", // FIXME: why inconsistent usage of _id vs. id?
```
nlarew, line 55: We _could_ switch `Message` to use `_id` but it'd need some prod db migration + code cleanup. Loosely I think `_id` implies the root-level document ID but I suppose it could also make sense in an embedded document.

nlarew (approved):
LGTM, couple small tweaks at your discretion

PR #188, opened 2023-09-26
Replace GitHub data source

cbush:
Jira: none

## Changes

- Uses makeGitDataSource as a drop-in replacement for GitHubRepoLoader.

## Notes

- Langchain's GitHubRepoLoader consumes a lot of API requests and quickly hits the limit. This disrupts our tests. It's particularly annoying given that we only load public repos.
- Will probably need special work to handle private repos.

### DIFF
--- a/ingest/.release-it.json
+++ b/ingest/.release-it.json
@@ -1,7 +1,9 @@
 {
   "git": {
+    "add": "../",
     "tag": false,
     "push": true,
+    "pushRepo": "upstream",
     "tagName": "ingest-v${version}"
   },
   "npm": {
@@ -13,6 +15,11 @@
     "release": true
   },
   "hooks": {
-    "before:init": ["git pull", "npm run build", "npm run lint", "npm run test"]
+    "before:init": [
+      "pushd ../chat-core && npm i && npm run build && popd",
+      "npm run build",
+      "npm run lint",
+      "npm run test"
+    ]
   }
 }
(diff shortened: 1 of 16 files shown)


ingest/src/GitHubDataSource.ts
```diff
@@ -47,53 +48,44 @@ export const makeGitHubDataSource = async ({
...
+  const { branch, recursive } = repoLoaderOptions ?? {};
+  if (branch) {
+    repoOptions["--branch"] = branch;
+  }
+  if (recursive) {
+    repoOptions["--recursive"] = `${recursive}`;
+  }
+
+  // Langchain GitHubRepoLoader consumes a lot of API requests and quickly hits
+  // the limit. This is now using makeGitDataSource as a drop-in replacement.
```
nlarew, line 61: Do we need the context comment?

nlarew (approved):

ingest/.release-it.json
```diff
@@ -1,7 +1,9 @@
 {
   "git": {
+    "add": "../",
```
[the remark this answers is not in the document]
cbush: Ok, I'm being _gaslit_ because the other day I was sure this was documented and working!

Replaced with the commitArgs -a

ingest/src/GitHubDataSource.ts
cbush: Prob not!

cbush: Actually, I think I will leave it because it really does seem weird that we have this source with a weird API just to wrap another source

PR #189, opened 2023-09-27
[DOCSP-32954] Add ErrorBanner, React-Transition-Group #182

jackyw2018:
### Resources

- [Jira Ticket](https://jira.mongodb.org/browse/DOCSP-32954)


chat-ui/src/DevCenterChatbot.tsx
```diff
@@ -41,11 +46,26 @@ const styles = {
       right: 49px;
     }
   `,
-  message_prompts: css`
+  message_prompts: (duration: number) => css`
```
nlarew, line 49: Are you planning on having multiple durations? If not, instead of using a function we could move `duration` up to be a root-level variable.

chat-ui/src/DevCenterChatbot.tsx
```diff
@@ -164,10 +180,17 @@ type InnerChatbotProps = {
 const WELCOME_MESSAGE =
   "Welcome to MongoDB AI Assistant. What can I help you with?";
 
-export function InnerChatbot({
+const SUGGESTED_PROMPTS = [
+  "How do you deploy a free cluster in Atlas?",
+  "How do you import or migrate data into MongoDB Atlas?",
+  "How do I get started with MongoDB?",
+  "Why should I use Atlas Search?",
+];
```
nlarew, line 188: For context, we now take these in as a prop on the docs chatbot. I think it's fine to hard code these for now but in our unified interface we'll probably want to use the props approach for both.

chat-ui/src/DevCenterChatbot.tsx
```diff
@@ -526,29 +566,59 @@ const Message = ({
...
+      in={inProp}
+      timeout={duration}
+      nodeRef={nodeRef}
+      classNames={styles.message_prompts(duration)}
+    >
+      <div className={styles.message_prompts(duration)} ref={nodeRef}>
+        <LGMessagePrompts label="Suggested Prompts">
+          {messagePrompts.map((sp, idx) => (
+            <MessagePrompt
+              key={idx}
```
nlarew, line 603:
```suggestion
              key={sp}
```

In general you shouldn't use the map index as a key - note the second [Rule of Keys](https://react.dev/learn/rendering-lists#rules-of-keys)

> Keys must not change or that defeats their purpose! **Don’t generate them while rendering.**

We have a much better candidate key in the suggested prompt text itself - each one is unique and the text won't change between renders.

nlarew (approved):
A couple small suggestions but overall LGTM!

PR #190, opened 2023-09-27
(DOCSP-31104) [UI] Add tests to FE services

nlarew:
Jira: (DOCSP-31104) [UI] Add tests to FE services

## Changes

- Adds tests for chat-ui services

## Notes

- No tests yet for the UI code itself (e.g. via react-testing-library) due to some unresolved ESM integration issues

### DIFF
--- a/chat-ui/src/services/conversations.test.ts
+++ b/chat-ui/src/services/conversations.test.ts
@@ -1,0 +1,261 @@
+import { vi } from "vitest";
+import { ConversationService, formatReferences } from "./conversations";
+import { References } from "chat-core";
+// import { MockEvent, EventSource } from "mocksse";
+import * as FetchEventSource from "@microsoft/fetch-event-source";
+
+/**
+ * Mocks
+ */
+
+/* Mock fetch for regular awaited HTTP requests */
+global.fetch = vi.fn();
+function mockFetchResponse<T = unknown>({
+  status = 200,
+  data,
+}: {
+  status?: number;
+  data: T;
+}) {
+  (global.fetch as ReturnType<typeof vi.fn>).mockResolvedValue({
+    status,
+    json: () => new Promise((resolve) => resolve(data)),
+  });
+}
+
+/* Mock @microsoft/fetch-event-source for SSE streaming requests */
+vi.mock("@microsoft/fetch-event-source");
+declare module "@microsoft/fetch-event-source" {
+  export type MockEvent = {
+    id?: string;
+    type?: string;
+    data: string | Record<string, unknown>;
+  };
+  export function __addMockEvents(...newEvents: MockEvent[]): void;
+  export function __clearMockEvents(): void;
+}
+
+/**
+ * Tests
+ */
+const serverUrl = "https://example.com/api/v1";
+// const serverUrl = "https://knowledge.staging.corp.mongodb.com/api/v1";
+
+describe("ConversationService", () => {
+  let conversationService: ConversationService;
+  beforeEach(() => {
+    conversationService = new ConversationService({
+      serverUrl,
+    });
+  });
+
+  it("throws if serverUrl is not set", () => {
+    expect(
+      () =>
+        new ConversationService({
+          serverUrl: undefined as unknown as string,
+        })
+    ).toThrow("You must define a serverUrl for the ConversationService");
+  });
+
+  it("creates conversations", async () => {
+    const conversationId = "650b4b260f975ef031016c8a";
+    mockFetchResponse({
+      data: {
+        _id: conversationId,
+        messages: [],
+      },
+    });
+    const conversation = await conversationService.createConversation();
+    expect(conversation.conversationId).toEqual(conversationId);
+    expect(conversation.messages).toEqual([]);
+  });
+
+  it("adds messages w/ an awaited response", async () => {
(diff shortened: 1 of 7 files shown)


nlarew:
Tests failing on ingest - likely solved once https://github.com/mongodb/chatbot/pull/188 merges and we can rebase this

chat-ui/src/services/conversations.test.ts
```diff
@@ -0,0 +1,261 @@
+import { vi } from "vitest";
+import { ConversationService, formatReferences } from "./conversations";
+import { References } from "chat-core";
+// import { MockEvent, EventSource } from "mocksse";
```
cbush, line 4: Needed?

chat-ui/src/services/conversations.test.ts
```diff
@@ -0,0 +1,261 @@
...
+import * as FetchEventSource from "@microsoft/fetch-event-source";
+
+/**
+ * Mocks
```
cbush, line 8: This is JSDoc notation. Just use normal comments `/*  ...  */` or `// ...`

chat-ui/src/services/conversations.test.ts
```diff
@@ -0,0 +1,261 @@
...
+ */
+
+/* Mock fetch for regular awaited HTTP requests */
```
cbush, line 11: Prefer line comments `//` - block comments are annoying when you're working and want to block comment something out
```suggestion
// Mock fetch for regular awaited HTTP requests
```

chat-ui/src/services/conversations.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  };
+  export function __addMockEvents(...newEvents: MockEvent[]): void;
+  export function __clearMockEvents(): void;
+}
+
+/**
+ * Tests
+ */
+const serverUrl = "https://example.com/api/v1";
+// const serverUrl = "https://knowledge.staging.corp.mongodb.com/api/v1";
```
cbush, line 42: Needed?

chat-ui/src/services/conversations.ts
```diff
@@ -59,10 +51,17 @@ export class TimeoutError<Data extends object = object> extends Error {
   }
 }
 
+type ConversationServiceConfig = {
```
cbush, line 54: Good idea to just export any type that is used in an exported public interface - in case users want to pre-construct the type or whatever
```suggestion
export type ConversationServiceConfig = {
```

chat-ui/src/services/conversations.ts
```diff
@@ -59,10 +51,17 @@ export class TimeoutError<Data extends object = object> extends Error {
...
+  serverUrl: string;
+};
+
 export class ConversationService {
   private serverUrl: string;
 
   constructor(config: ConversationServiceConfig) {
+    if(!config.serverUrl) {
```
cbush, line 62: Prettier malfunction?
```suggestion
    if (!config.serverUrl) {
```

chat-ui/src/services/conversations.ts
```diff
@@ -59,10 +51,17 @@ export class TimeoutError<Data extends object = object> extends Error {
...
+      throw new Error("You must define a serverUrl for the ConversationService");
```
cbush, line 63: This looks like a spot where `assert()` would be more appropriate (programmer error/invariant failure)

chat-ui/src/services/conversations.ts
```diff
@@ -292,14 +291,23 @@ export class ConversationService {
...
     });
+
+    if(res.status === 204) {
+      return rating;
+    }
+    if (res.status >= 400) {
+      const data = await res.json();
+      throw new Error(data.error);
+    }
+    throw new Error(`Server error: ${res.statusText ?? res.status}`);
```
cbush, line 311: Is it necessarily an error here? We could say
```suggestion
    throw new Error(`Server returned unexpected status ${res.status} (message: ${res.statusText})`);
```
(Will render as 'undefined' for undefined status)

chat-ui/vitest.config.ts
```diff
@@ -0,0 +1,38 @@
...
+const testConfig = defineConfig({
+  test: {
+    environment: "jsdom",
+    globals: true,
+    // alias: [
+    //   {
+    //     find: new RegExp("^@lg-chat/message$", "g"),
+    //     replacement: "@lg-chat/message/dist/index.js",
+    //   },
+    // ],
```
cbush, line 13: What's this for?

cbush (changes requested):
Looks good and functional so far, but I'm mildly bothered by enough superficial things that I'd like to take another look before we merge

chat-ui/vitest.config.ts
nlarew: Leftover debugging of ESM issues when testing the components. `react-markdown` (used in `Message`) causes ESM import issues in test so this was trying to work around it.

chat-ui/src/services/conversations.ts
nlarew: 5XX errors _do_ imply some kind of server error but the error message doesn't necessarily need to say that verbatim

cbush (approved):
Thanks!

PR #191, opened 2023-09-29
Add whether message was deemed "Do not answer" to conversation

cbush:
Jira: none

## Changes

- Refactors Messages to use a more TypeScript-y pattern (even though it is somewhat more annoying to use)

### DIFF
--- a/scripts/src/scrubMessages.ts
+++ b/scripts/src/scrubMessages.ts
@@ -1,5 +1,11 @@
 import { MongoClient, Db, Collection, Document } from "mongodb";
-import { Conversation } from "chat-server";
+import {
+  Conversation,
+  Message,
+  UserMessage,
+  AssistantMessage,
+  SystemMessage,
+} from "chat-server";
 import { ScrubbedMessage } from "./ScrubbedMessage";
 
 import "dotenv/config";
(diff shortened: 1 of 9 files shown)


scripts/src/scrubMessages.ts
```diff
@@ -61,7 +67,26 @@ const scrubMessages = async ({ db }: { db: Db }) => {
...
+          // just use ScrubbedMessage/SomeMessage because we want the union of
+          // all possible Message-type keys.
+          | keyof UserMessage
+          | keyof AssistantMessage
+          | keyof SystemMessage
+          | keyof ScrubbedMessage,
+          // Add keys to omit from the projection below.
+          | "id"
+          | "content"
+          | "preprocessedContent"
```
nlarew, line 84: Are these not already included? I figured since the all the `FooMessage` types extend `Message` these would already be in the union.

nlarew (approved):
LGTM + one question

scripts/src/scrubMessages.ts
cbush: We're explicitly trying to exclude these fields here so they don't get projected into the scrubbed version, if that makes sense.

cbush: Of course they won't get projected as-is, but by adding it to the `requires` clause we encode the intention to keep them out.

PR #193, opened 2023-10-02
[DOCSP-32950] Create a DevCenter Chatbot Component

jackyw2018:
Jira: https://jira.mongodb.org/browse/DOCSP-32950

## Changes

- Add a DevCenter Chatbot Component, which is a Floating Action Button that triggers a Chatbot Modal


nlarew (changes requested):
A couple small changes but then this should be good to go!

nlarew (approved):

PR #199, opened 2023-10-06
Conversations refactor

cbush:
Jira: none

## Changes

- Increases code readability and enable further refactors by refactoring makeAddMessageToConversationRoute to centralize error handling under one catch block
- Simplifies implementation and improves extensibility by baking sprawled dependencies into a single injected `findContent` function
- Defines `findContent` in configuration

## Notes

- TODO? Apply similar refactor principles to other paths

### DIFF
--- a/chat-server/src/routes/conversations/addMessageToConversation.ts
+++ b/chat-server/src/routes/conversations/addMessageToConversation.ts
@@ -1,7 +1,6 @@
 import {
   Request as ExpressRequest,
   Response as ExpressResponse,
-  NextFunction,
 } from "express";
 import {
   ObjectId,
@@ -11,10 +10,8 @@ import {
   FindNearestNeighborsOptions,
   References,
   Reference,
-  WithScore,
 } from "chat-core";
 import {
-  Conversation,
   ConversationsService,
   Message,
   conversationConstants,
@@ -74,6 +71,15 @@ export interface AddMessageToConversationRouteParams {
   userQueryPreprocessor?: QueryPreprocessorFunc;
   maxChunkContextTokens?: number;
 }
+
+export type RequestError = Error & {
+  httpStatus: number;
+};
+
+export const makeRequestError = (
+  args: Omit<RequestError, "name">
+): RequestError => ({ ...args, name: "RequestError" });
+
 export function makeAddMessageToConversationRoute({
   store,
   conversations,
@@ -85,11 +91,7 @@ export function makeAddMessageToConversationRoute({
   userQueryPreprocessor,
   maxChunkContextTokens = 1500,
 }: AddMessageToConversationRouteParams) {
-  return async (
-    req: ExpressRequest,
-    res: ExpressResponse<ApiMessage>,
-    next: NextFunction
-  ) => {
+  return async (req: ExpressRequest, res: ExpressResponse<ApiMessage>) => {
     const reqId = getRequestId(req);
     try {
       const {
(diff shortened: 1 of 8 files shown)


chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -101,108 +101,65 @@ export function makeAddMessageToConversationRoute({
...
+        throw makeRequestError({
           httpStatus: 400,
-          errorMessage: `Invalid IP address ${ip}`,
+          message: `Invalid IP address ${ip}`,
         });
       }
 
-      const shouldStream = Boolean(stream);
-      const latestMessageText = message;
+      const latestMessageText = message as string;
```
nlarew, line 117: Why the cast to string here?

cbush: Holdover from ExpressRequest not being properly typed. Let me see if I can fix that...

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -357,64 +296,70 @@ export function makeAddMessageToConversationRoute({
...
-        dataStreamer.streamData({
-          type: "finished",
-          data: apiRes.id,
-        });
-        dataStreamer.disconnect();
-        return;
-      } else {
-        res.status(200).json(apiRes);
+
+      if (!shouldStream) {
```
mongodben, line 327: should this be an `else` statement instead of second `if` statement?

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -101,108 +105,65 @@ export function makeAddMessageToConversationRoute({
...
-          httpStatus: 400,
-          errorMessage: "Invalid conversation ID",
-        });
-      }
-
       if (!isValidIp(ip)) {
-        return sendErrorResponse({
-          reqId,
-          res,
+        throw makeRequestError({
```
mongodben, line 115: totally didn't understand til reading this PR that you can throw any object. thought it needed to be using the `Error` class.

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -357,64 +296,70 @@ export function makeAddMessageToConversationRoute({
...
+      if (!shouldStream) {
+        return res.status(200).json(apiRes);
       }
-    } catch (err) {
-      logRequest({
-        reqId,
-        message: "An unexpected error occurred: " + JSON.stringify(err),
-        type: "error",
+
+      dataStreamer.streamData({
```
mongodben, line 331: i think this stuff could be inside this `if` statement https://github.com/mongodb/chatbot/pull/199/files#diff-ad16696d33139a2879677f2ca6d0ef3291daf4386dc13ecfad6b735809392134R261

mongodben (changes requested):
changes look great. much clearer. this route had evolved into something of a frankenstein. 1 comment about data streamer. 

also see logs of test fail. not sure what's going on there - https://drone.corp.mongodb.com/mongodb/chatbot/1527/1/2

also, have you tested out that stuff is still working in dev env end-to-end with this change using npmv run dev from root of project? given that there's a lot of refactoring happening here, would be good to do quick manual validation

chat-server/src/routes/conversations/addMessageToConversation.ts
cbush: We need the apiRes.id which is only generated after storing the message in the database. This could probably be refactored further but I'd rather iterate on this PR later.

mongodben (approved):
LGTM!

like we talked about during standup, can you create a new ticket to add the query preprocessor to the `FindContentFunc`?

would be good to have that in there before we OSS the thing

mongodben:
> * TODO? Apply similar refactor principles to other paths

def worth a look. they should be much simpler to refactor b/c their logic is much simpler. 

could you make a ticket for this as well?

cbush:
Done: https://jira.mongodb.org/browse/DOCSP-33718

PR #205, opened 2023-10-16
(DOCSP-32603): Add Terraform data source

mongodben:
Jira: https://jira.mongodb.org/browse/DOCSP-32603

## Changes

- Add Terraform data source

## Notes

-

### DIFF
--- a/chat-core/src/extractFrontMatter.ts
+++ b/chat-core/src/extractFrontMatter.ts
@@ -1,15 +1,15 @@
 import * as matter from "gray-matter";
 import toml from "toml";
 
-export function extractFrontMatter(
+export function extractFrontMatter<T = Record<string, unknown>>(
   text: string,
   language?: string,
   delimiter?: string
 ): {
-  metadata?: Record<string, unknown>;
+  metadata?: T;
   body: string;
 } {
-  let metadata: Record<string, unknown> | undefined;
+  let metadata: T | undefined;
   let body = text;
   const options = {
     delimiters: delimiter,
@@ -23,7 +23,7 @@ export function extractFrontMatter(
     ? matter.default(text, options)
     : undefined;
   body = frontmatterResult?.content ?? text;
-  metadata = frontmatterResult?.data;
+  metadata = frontmatterResult?.data as T | undefined;
 
   return {
     metadata,
--- a/ingest/src/projectSources.ts
+++ b/ingest/src/projectSources.ts
@@ -14,6 +14,8 @@ import {
   MakeMdOnGithubDataSourceParams,
   makeMdOnGithubDataSource,
 } from "./MdOnGithubDataSource";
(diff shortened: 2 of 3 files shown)


chat-core/src/extractFrontMatter.ts
```diff
@@ -1,15 +1,15 @@
 import * as matter from "gray-matter";
 import toml from "toml";
 
-export function extractFrontMatter(
+export function extractFrontMatter<T = Record<string, unknown>>(
```
cbush, line 4: I doubt we want this to be any T (number, string, ...) but rather any object?
```suggestion
export function extractFrontMatter<T extends Record<string, unknown> = Record<string, unknown>>(
```

ingest/src/projectSources.ts
```diff
@@ -504,6 +506,53 @@ const practicalAggregationsDataSource = async () => {
...
+            .replace(".markdown", "");
+      } else if (path.includes("website/docs/r/")) {
+        url =
+          siteBaseUrl +
+          path.replace("website/docs/r", "resources").replace(".markdown", "");
+      } else {
+        url =
+          siteBaseUrl +
+          path.replace("website/docs/", "").replace(".markdown", "");
+      }
```
cbush, line 544: The construction 
```
let temporarilyUndefinedVariable;
if (...) {
  temporarilyUndefinedVariable = thiz
} else if (...) {
  temporarilyUndefinedVariable = that  
} ...
```

is a pretty good indicator for refactor.

```js
const url = getUrlForBlahblah(path);
...
function getUrlForBlahblah(url) {
  if (...) {
    return thiz
  } else if (...) { ...
    return that
  } ...
}
```

This will actually also highlight dependency sprawl (think: opportunity for merging into a single findContentFunc style thing?) as it forces you to move scope variables into a parameter list for the helper function.

Btw, I would not export or explicitly test the helper function (`getUrlForBlahblah()` in this example).

ingest/src/projectSources.ts
```diff
@@ -504,6 +506,53 @@ const practicalAggregationsDataSource = async () => {
...
+      "--branch": "master",
+    },
+    metadata: {
+      productName: "mongodbatlas Terraform Provider",
+      tags: ["docs", "terraform", "atlas", "hcl"],
+    },
+    filter: (path: string) =>
+      path.includes("website/docs") && path.endsWith(".markdown"),
+    handlePage: async function (path, content) {
+      const { metadata, body } = extractFrontMatter<{ page_title: string }>(
```
cbush, line 526: page_title: have we used this case style before? Any particular reason for introducing it here?

cbush (changes requested):
A few comments

ingest/src/projectSources.ts
mongodben: this comes from the frontmatter in the markdown file we're ingesting. not something we're defining

PR #206, opened 2023-10-16
OSS Chat Server

mongodben:
Jira: https://jira.mongodb.org/browse/DOCSP-33726

## Changes

Make the following changes to the `chat-server` and `chat-core` code to make the server more usable by other teams in their projects:

- [x] Add TSDoc comments throughout the project
- [x] Publish the chat-server package as an npm module
- [x] Refactor `chat-core/src/OpenAiEmbedFunc.ts` to use `@azure/openai` SDK
- [x] Create separate `makePageStore` and `makeEmbeddedContentStore` constructor functions out of `makeDatabaseConnection`.
- [x] Remove unnecessary `MongoDbInterface` abstraction
- [x] Make server base URL configurable
- [x] Make max number messages & tokens in single RAG in conversation configurable
- [x] Make the creation of reference links from the data returned from the `EmbeddedContentStore` configurable. 

## Notes

Next steps:
- As a next step as part of the OSS-ification, we really should split our implementation and the stuff specific to our implementation in a separate npm package. We need this b/c we need to version our implementation of chatbot for publishing to prod, and need to version the package.
  - Also, there's some MongoDB-specific stuff in the package that doesn't need to be, like our preprocessor. 
  - Improve library package publishing workflow as part of this
  - new ticket created for this: https://jira.mongodb.org/browse/DOCSP-33754
- Publishing the chat server doesn't work. ticket to do this - https://jira.mongodb.org/browse/DOCSP-33823


ingest/src/commands/all.ts
```diff
@@ -18,20 +19,25 @@ const commandModule: CommandModule<unknown, unknown> = {
     const { MONGODB_CONNECTION_URI, MONGODB_DATABASE_NAME } =
       assertEnvVars(INGEST_ENV_VARS);
 
-    const store = await makeDatabaseConnection({
+    const embeddedContentStore = makeMongoDbEmbeddedContentStore({
```
mongodben, line 22: catch both these

chat-core/src/EmbeddedContent.ts
```diff
@@ -102,7 +111,7 @@ export type FindNearestNeighborsOptions = {
   minScore: number;
 
   /**
-    Atlas Search filter expression.
+    Search filter expression.
```
cbush, line 114: Heh, this does highlight a bias (obviously) to Atlas VS, because the user of the abstract interface will need to know some implementation details in order to provide the right kind of search filter.

Nothing to do now, just something worth keeping in mind for later.

chat-core/src/MongoDbDatabaseConnection.ts
```diff
@@ -0,0 +1,43 @@
...
+  db: Db;
+} {
+  const mongoClient = new MongoClient(connectionUri, {
+    serverSelectionTimeoutMS: 30000,
+  });
+  const db = mongoClient.db(databaseName);
+  return {
+    mongoClient,
+    db,
+    async drop() {
```
cbush, line 27: I still think this should have been called something like `debugDropDatabase()` just to really make clear that this should not be used in non-test code.

Actually, we might want to refactor this out and have the tests instead use MongoDB directly to drop the test databases.

```js
makeTestMongoDbDatabaseConnection(...) {
  const connection = makeMongoDbDatabaseConnection(...);
  return { ...connection, drop() { /* implement drop logic here */ } };
}

File: chat-core/src/MongoDbDatabaseConnection.ts [hunk 4]
```diff
@@ -0,0 +1,43 @@
...
+    async close(force) {
+      mongoClient.close(force);
+    },
+  };
+}
+
+/**
+  Returns a query filter that represents a unique page in the system.
+ */
+export const pageIdentity = ({ url, sourceName }: Page) => ({
```
cbush, line 40: Does this belong in this file at all?

chat-core/src/MongoDbEmbeddedContentStore.ts
```diff
@@ -0,0 +1,125 @@
...
+  makeMongoDbDatabaseConnection,
+  pageIdentity,
+} from "./MongoDbDatabaseConnection";
+import { strict as assert } from "assert";
+
+export function makeMongoDbEmbeddedContentStore({
+  connectionUri,
+  databaseName,
+}: MakeMongoDbDatabaseConnectionParams): EmbeddedContentStore &
+  DatabaseConnection {
```
cbush, line 19: This might be useful for testing, so I'm not saying to remove it, but what does `DatabaseConnection` add here that `EmbeddedContentStore` doesn't have? Maybe `close()` (in which case - probably the Stores should have a `close()` function)

chat-core/src/MongoDbPageStore.ts
```diff
@@ -0,0 +1,55 @@
+import { DatabaseConnection } from "./DatabaseConnection";
+import {
+  MakeMongoDbDatabaseConnectionParams,
+  makeMongoDbDatabaseConnection,
+  pageIdentity,
```
cbush, line 5: Might be local only (not exported) from this file

chat-core/src/Page.ts
```diff
@@ -73,3 +80,8 @@ export type PageStore = {
    */
   updatePages(pages: PersistedPage[]): Promise<void>;
 };
+
+/**
+  OSS_TODO: per comment on DatabaseConnection.ts, i think we should make a separate
+  makePageStore() func here.
```
cbush, line 86: Still needed?

cbush (changes requested):
Looks great, but will take another look

chat-core/src/Page.ts
mongodben: nay

chat-core/src/EmbeddedContent.ts
mongodben: good point! 

i think `Record<string, unknown>` is generic enough that it could apply to other vector stores as well, but perhaps we should make this a passable `T` to the method or a constructor func

chat-core/src/MongoDbDatabaseConnection.ts
mongodben: i think i'll just note in the TSDoc comment `@internal`

chat-core/src/MongoDbDatabaseConnection.ts
mongodben: will move to own file

chat-core/src/MongoDbEmbeddedContentStore.ts
mongodben: i get what you're saying. will add it as an optional aspect b/c i can see not all stores needing a `.close()` method

chat-core/src/MongoDbPageStore.ts
mongodben: it's used in 2 files. making it it's own file

examples/basic-chatbot-server/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "target": "es6",
+    "noImplicitAny": true,
+    "moduleResolution": "node",
+    "sourceMap": true,
+    "outDir": "build",
+    "baseUrl": ".",
+    "paths": {
+      "*": ["node_modules/*", "src/types/*"]
+    }
+    // TODO: how to compile dependencies?
```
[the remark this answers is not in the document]
mongodben: i don't recall what the intent of this comment was. removing

chat-core/src/MongoDbEmbeddedContentStore.ts
cbush: At first I wasn't sure but I think I like that!

Optional `cleanUp()` on every store?

chat-core/src/MongoDbDatabaseConnection.ts
cbush: I'm not 100% satisfied but this isn't something we have to fix today. If someone is calling drop() in production code then...

chat-core/src/EmbeddedContent.ts
cbush: `Record<string, unknown>` is fine but the problem is that the exact shape of that object (think MQL) depends on the actual backing vector store.

PR #207, opened 2023-10-17
Organize some files

cbush:


ingest/src/commands/all.test.ts
```diff
@@ -1,9 +1,7 @@
 import { PageStore, EmbeddedContentStore, assertEnvVars } from "chat-core";
 import { MongoClient } from "mongodb";
-import "dotenv/config";
-import { INGEST_ENV_VARS } from "../IngestEnvVars";
 import { doAllCommand } from "./all";
-import { makeIngestMetaStore } from "../IngestMetaStore";
+import { INGEST_ENV_VARS, makeIngestMetaStore } from "../";
```
mongodben, line 4: should we not import from the `index.ts` files b/c it can create circular dependencies?

ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -1,7 +1,7 @@
 import { EmbeddedContentStore, EmbeddedContent, Page } from "chat-core";
 import { updateEmbeddedContent } from "./updateEmbeddedContent";
-import { makeMockPageStore } from "./updatePages.test";
-import { persistPages } from "./updatePages";
+import { makeMockPageStore } from "../pages/updatePages.test";
```
mongodben, line 3: maybe we can make a testUtils file for this instead of importing from 1 test to another?

mongodben (commented):
couple small comments

ingest/src/embed/updateEmbeddedContent.test.ts
cbush: I think "Utils" is an anti-pattern, but yes, let's see

ingest/src/commands/all.test.ts
cbush: Yes, rethought this greatly. Not just for circular dependencies but also as an anti-pattern. Thanks!

ingest/src/commands/all.ts
```diff
@@ -8,9 +8,9 @@ import {
   PageStore,
   logger,
 } from "chat-core";
+import { INGEST_ENV_VARS, makeIngestMetaStore } from "..";
```
mongodben, line 11: still index import here

ingest/src/sources/GitHubDataSource.ts
```diff
@@ -1,9 +1,6 @@
 import { Document } from "langchain/document";
-import {
-  GithubRepoLoader,
-  GithubRepoLoaderParams,
-} from "langchain/document_loaders/web/github";
-import { Page, logger } from "chat-core";
+import { GithubRepoLoaderParams } from "langchain/document_loaders/web/github";
```
mongodben, line 2: (not in this PR) we should refactor this to just use the Git data source

ingest/src/sources/snooty/SnootyProjectsInfo.ts
```diff
@@ -5,8 +5,7 @@ import {
   makeSnootyDataSource,
   Branch,
 } from "./SnootyDataSource";
-import { LocallySpecifiedSnootyProjectConfig } from "./projectSources";
-import { DataSource } from "./DataSource";
+import { DataSource, LocallySpecifiedSnootyProjectConfig } from "..";
```
mongodben, line 8: fix index import

mongodben (approved):
a few index imports to fix, but otherwise lgtm!

ingest/src/sources/GitHubDataSource.ts
cbush: We did!

cbush: What we should do is delete the GitHub thing entirely.

PR #209, opened 2023-10-18
Api chat

mongodben:
for skunkworks

- architecture overview
- sketch of types and methods


api-chat.md
```diff
@@ -0,0 +1,57 @@
...
+
+## TODOs
+
+Stuff to do in project. As week goes on, we'll probably want to break this stuff down further.
+
+- [ ] [INGEST] Ingest API spec files to `ApiEmbeddedContent` MongoDB collection
+- [ ] [MONGODB] Set up Atlas Vector Search on `ApiEmbeddedContent` collection
+- [ ] [UI] Make so can load API key, Atlas project and group before starting to chat
+  - this'll make the flow work better.
+  - send these with each req to server
```
nlarew, line 15: How do we see this working? e.g.

1. User says "please add a search index to my collection"
2. Assistant responds with a function call description + text inputs for the API key
3. User enters their API key and then clicks "confirm please do this"
4. We call the function on the server with an RPC API

I think this is fine for skunkworks but should motivate a better way to do this longer term. i.e. when the user is logged in (would need to add SSO to docs) we can use that to connect "over the mesh" like app services apps do. IIUC this lets us connect to and administrate the user's cluster without creating db users or needing API keys (citation needed).

mongodben: as a "step 0", i was think we just have the user add their API key, project ID, and cluster ID to the UI.

in a real atlas chat app this would be handled under the hood, something along the lines of what you described above. but for skunk just passing the values should be fine.

PR #211, opened 2023-10-20
Ingest plugins

cbush:
Jira: https://jira.mongodb.org/browse/DOCSP-33727

## Changes

- Implement config loader. Every config property can be either an instance or a sync/async constructor of an instance. This allows users to instantiate asynchronously.
- Config uses flat config style of eslintrc.
- Create ingest-mongodb-public npm project that defines the standard MongoDB docs chatbot data sources config
- Move everything configurable from CLI command files to config
- Change "EmbedFunc" function to "Embedder" object (with `embed()` method) - this allows us to differentiate between a constructor and a constructed embed func in the config object. We can also decorate the object with more properties later if needed (ignore that technically you can do this with a function too in JS)

## Notes

- ingest-mongodb-public will need to be the new ingest on cron

## TODO

- [x] Implement config cleanup call in the finally blocks of `withConfig()`
- [x] Update cron to use ingest-mongodb-public instead of ingest
- [x] Fill out README for ingest-mongodb-public 
- [x] Handle bad config

## Future Scope

- Allow specifying config by git repo URL etc.


ingest-mongodb-public/README.md
nlarew, line 1: This page intentionally left blank?

ingest-mongodb-public/environments/production.yml
```diff
@@ -0,0 +1,36 @@
+image:
+  repository: node
+  tag: 18
+
+cronJobs:
+  - name: ingest-all-content
+    schedule: "0 6 * * *" # every day at 6am UTC (1AM EST)
+    command: ["node", ".", "all"]
```
nlarew, line 8: Will this still work given that this is no longer in `ingest`? Do we need to update/move the `ingest-service.dockerfile` to this project?

cbush: Yep, good catch. Need to do this. Will do next

mongodben: ticket for this pls

chat-core/src/arrayFilters.ts
```diff
@@ -0,0 +1,15 @@
+/**
+  Given an array of PromiseSettledResults, return those that are fulfilled as PromiseFulfilledResults.
+ */
+export const filterFulfilled = <T>(
```
mongodben, line 4: nice util

chat-core/src/Embedder.ts
```diff
@@ -7,4 +7,6 @@ export type EmbedResult = {
   embedding: number[];
 };
 
-export type EmbedFunc = (args: EmbedArgs) => Promise<EmbedResult>;
+export type Embedder = {
```
mongodben, line 10: can you give this a little TSdoc?

ingest/package.json
mongodben, line 1: on line 2, we should give package a new name instead of "ingest". something like "mongodb-rag-ingest-cli" or smthn. 

note that i don't think we should have "chatbot" (even though it's in the server and ui package names) in the ingest package name, b/c our ingestion is more general than chatbot usecase

ingest-mongodb-public/environments/production.yml
```diff
@@ -5,7 +5,14 @@ image:
...
     schedule: "0 6 * * *" # every day at 6am UTC (1AM EST)
-    command: ["node", ".", "all"]
+    command:
+      [
+        "node",
+        "./node_modules/ingest/build/main.js",
+        "all",
+        "--config",
+        "./build/config.js",
+      ]
```
mongodben, line 15: maybe instead we can make a package.json script for this. smthn like: 

```shell
npm run ingest:all
```

and that script would look smthn like:

```json
"ingest:all": "mongodb-rag-ingest-cli all --config ./build/config.js"
```

ingest-mongodb-public/src/config.ts
```diff
@@ -0,0 +1,63 @@
...
+  databaseName: MONGODB_DATABASE_NAME,
+});
+
+const ingestMetaStore = makeIngestMetaStore({
+  connectionUri: MONGODB_CONNECTION_URI,
+  databaseName: MONGODB_DATABASE_NAME,
+  entryId: "all",
+});
+
+const pageStore = embeddedContentStore;
```
mongodben, line 42: this will change based on the changes i made to split embedded content and pages stores i think

ingest-mongodb-public/src/sources.ts
```diff
@@ -1,38 +1,25 @@
...
+} from "ingest/sources";
+import {
+  prepareSnootySources,
+  LocallySpecifiedSnootyProjectConfig,
+} from "ingest/sources/snooty";
 
 /**
   Async constructor for specific data sources -- parameters baked in.
  */
 export type SourceConstructor = () => Promise<DataSource | DataSource[]>;
```
mongodben, line 21: should this be in the "ingest" package?

ingest/src/sources/snooty/SnootyDataSource.ts
```diff
@@ -75,6 +75,19 @@ export type SnootyProjectConfig = ProjectBase & {
   baseUrl: string;
 };
 
+// `baseUrl` and `currentBranch` to be filled in by the Snooty Data API GET
```
mongodben, line 78: should this be TSDoc?

ingest/src/withConfig.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const configArray = (await import(path)).default as Config[];
+
+  // Validate config
+  if (!Array.isArray(configArray)) {
+    throw new Error(
+      `Invalid config at ${path}: expected default exported array of Config objects`
+    );
+  }
+  // TODO: further validation
```
mongodben, line 24: @nlarew did some nice type validation stuff on the server w zod. wonder if we could do same here.

ingest/src/withConfig.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  // Flatten config
```
mongodben, line 26: can you explain what is the behavior here with the flattening?

ingest/src/withConfig.ts
```diff
@@ -0,0 +1,88 @@
...
+) => {
+  const config = await loadConfig(args);
+  const resolvedConfig = await resolveConfig(config);
+
+  console.log(`resolved config: ${resolvedConfig.dataSources[0]}`);
+  try {
+    return await action(resolvedConfig, args);
+  } finally {
+    // TODO
+    // config.embeddedContentStore.close();
```
mongodben, line 47: all the stores need to be closed, right?

mongodben (commented):
did a quickish scan through. looks very promising and powerful!

ingest-mongodb-public/environments/production.yml
cbush: No? It's done as part of this PR

ingest-mongodb-public/src/config.ts
cbush: I know, which is why I needed that merged

ingest-mongodb-public/src/sources.ts
cbush: I don't think so. This is pretty implementation specific for snooty sources.

ingest/src/withConfig.ts
cbush: Correct

ingest-mongodb-public/README.md
```diff
@@ -0,0 +1,24 @@
...
+npm i
+npm run build
+```
+
+## Using the Config
+
+Run `ingest`, passing the built js file to the `--config` flag.
+
+```sh
+node ./node_modules/ingest/build/main.js \
```
mongodben, line 21: i think this can be updated with: `npm run ingest:all`

ingest-mongodb-public/src/sources.ts
mongodben: maybe rename `SnootySourceConstructor` then?

ingest-service.dockerfile
mongodben, line 1: have you tested that this builds, deploys to drone, and runs in k8s properly?

ingest/src/Config.ts
```diff
@@ -0,0 +1,51 @@
...
+  It also avoids unnecessary construction and cleanup if that field of the
+  config is overridden by a subsequent config.
+ */
+export type Config = {
+  /**
+    The store that contains the ingest meta document.
+
+    The ingest meta document is used by the `all` command to determine when the
+    last successful run was for that ingestion.
+   */
```
mongodben, line 22: i could envision us expanding the usage of the meta store for more than the all command. perhaps rephrase to be more generic to take into account future use cases?

ingest/src/Config.ts
```diff
@@ -0,0 +1,51 @@
...
+  ingestMetaStore: Constructor<IngestMetaStore>;
+
+  /**
+    The store that holds pages downloaded from data sources.
+   */
+  pageStore: Constructor<PageStore>;
+
+  /**
+    The store that holds the embedded content for later vector search.
```
mongodben, line 31:
```suggestion
    The store that holds the embedded content and vector embeddings for later vector search.
```

ingest/src/withConfig.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  if (configArray.length === 0) {
+    throw new Error(
+      `Invalid config at ${path}: expected at least 1 Config object in array`
+    );
+  }
+
+  // Flatten config: the fields of earlier entries in the array can be
+  // overridden by the fields of subsequent entries in the array.
```
mongodben, line 33: i don't think this is desired behavior. the way i think about it, each `Config` in the array should be independent. why do you think it should be this way?

mongodben (changes requested):
generally looks great. a few small things noted in comments, and also this one which might be a bit bigger https://github.com/mongodb/chatbot/pull/211/files#r1369079889

ingest-service.dockerfile
mongodben: if you need help testing the infra, it's prob easiest to do together in a zoom

ingest/src/withConfig.ts
cbush: That's how it works in eslint's flat cascade? https://eslint.org/blog/2022/08/new-config-system-part-2/#goodbye-extends%2C-hello-flat-cascade

mongodben: i think that factor of the flat config would only apply here if both configs touch the same data source. ie both are for `snooty-docs-realm`. but if there's config1 with `snooty-docs-realm` using embeddedContentStore1`, and config2 with `snooty-docs-app-services` using `embeddedContentStore2`, that should be allowed.

ingest-service.dockerfile
mongodben: per convo, we'll do this later

mongodben (approved):
LGTM! nice work here. think we ended up with a very solid and extensible config system

PR #220, opened 2023-10-23
adding a trailing slash to env variables expected to be URLs

nkulkarni:
## Changes

- add a trailing `/` to urls

## Notes

- before, if you didn't have a trailing `/` to the url of, say, the OpenAI base path, it got concatenated to the deployment ID without having a separator

### DIFF
--- a/chat-core/src/assertEnvVars.ts
+++ b/chat-core/src/assertEnvVars.ts
@@ -1,3 +1,6 @@
+import { URL_ENV_VARS } from "CoreEnvVars";
+
+
 export const assertEnvVars = <T extends Record<string, string>>(
   ENV_VARS: T
 ): T => {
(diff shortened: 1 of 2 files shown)


chat-core/src/assertEnvVars.ts
```diff
@@ -7,6 +10,11 @@ export const assertEnvVars = <T extends Record<string, string>>(
       // Get this var value from process.env - either valid string or undefined
       vars[currentVar] = process.env[currentVar];
 
+      // if the var is a url, make sure it has a following slash
+      if (URL_ENV_VARS.includes(currentVar)) {
```
cbush, line 14: assertEnvVars is agnostic to exact env var names (it is used it other parts). We would need another way of determining what the URLs are supposed to be.

Can we instead assert for trailing slashes where the vars are actually used?

chat-core/src/assertEnvVars.ts
```diff
@@ -21,3 +29,5 @@ ${Object.entries(vars)
   }
   return vars as T;
 };
+
+const ensure_url_slash_termination = (url: string): string => url.endsWith("/") ? url : `${url}/`;
```
cbush, line 33: Prefer camelCase

cbush (changes requested):
Thanks for the PR! We'd need to keep assertEnvVars generic, but it's a good issue to fix.

PR #221, opened 2023-10-23
(DOCSP-33348) [UI] Abstract chatbot component for multiple sites

nlarew:
Jira: (DOCSP-33348) [UI] Abstract chatbot component for multiple sites

## Changes

- Docs & DevCenter UI now share code for all common components
- Refactored Chatbot into a wrapper component (`<Chatbot>`) which provides data + trigger & view child components which share most chatbot state & data via context
  - Triggers are the default view & how you start interacting with the chatbot. For docs the trigger is an input bar, for devcenter it's a floating action button.
  - Views are the actual conversation. For now we just have the modal view but in future we could have a full screen view, fixed chat sidebar, embedded widget, etc.


jackyw2018 (approved):
LGTM!

PR #231, opened 2023-10-30
Script: sampleAndSaveMarkdownPages

nlarew:
Samples N documents from the `pages` collection and saves them as separate markdown files each containing one page's `body` content.


scripts/src/sampleAndSaveMarkdownPages.ts
```diff
@@ -0,0 +1,89 @@
...
+  return res;
+}
+
+function partition<T, Pass extends T = T, Fail extends T = T>(
+  arr: T[],
+  test: (elem: T) => boolean
+) {
+  type PassFail = [Pass[], Fail[]];
+  const passes: Pass[] = [];
+  const fails: Fail[] = [];
```
cbush, line 50: I think you create possible pain for yourself declaring these locally and having them shadowed within the reducer. I would just put `baseAccumulator = [[], []];`

cbush (approved):

PR #234, opened 2023-10-31
(DOCSP-33030) [Q&A] Record request origin for usage tracking

nlarew:
Jira: (DOCSP-33030) [Q&A] Record request origin for usage tracking

## Changes

- Requires and checks for the `Origin` HTTP header and saves it alongside every conversation & user message

Screenshot 2023-10-31 at 5 22 48 PM

### DIFF
--- a/chat-server/src/routes/conversations/addMessageToConversation.ts
+++ b/chat-server/src/routes/conversations/addMessageToConversation.ts
@@ -182,6 +182,7 @@ export function makeAddMessageToConversationRoute({
           dataStreamer,
           requestOrigin,
           res,
+          originalMessageEmbedding: [],
         });
       }
       const query = preprocessedUserMessageContent ?? latestMessageText;
@@ -206,6 +207,7 @@ export function makeAddMessageToConversationRoute({
           dataStreamer,
           requestOrigin,
           res,
+          originalMessageEmbedding: queryEmbedding,
         });
       }
 
@@ -364,6 +366,7 @@ export async function sendStaticNonResponse({
   res,
   rejectQuery,
   requestOrigin,
+  originalMessageEmbedding,
 }: {
   conversations: ConversationsService;
   conversation: Conversation;
@@ -373,7 +376,8 @@ export async function sendStaticNonResponse({
   shouldStream: boolean;
   dataStreamer: DataStreamer;
   res: ExpressResponse<ApiMessage>;
-  requestOrigin?: string;
+  requestOrigin: string;
+  originalMessageEmbedding: number[];
 }) {
   const { assistantMessage } = await addMessagesToDatabase({
     conversations,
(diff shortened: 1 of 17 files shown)


chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -95,11 +79,13 @@ export function makeAddMessageToConversationRoute({
     res: ExpressResponse<ApiMessage>
   ) => {
     const reqId = getRequestId(req);
+
     try {
       const {
         params: { conversationId: conversationIdString },
         body: { message },
         query: { stream },
+        headers: { origin: requestOrigin },
```
mongodben, line 88: could the logic around getting origin be middleware instead of baked into the route?

and then for persisting to the DB, i think we could make an injected function that takes the request object and adds arbitrary data from it (including origin) to the object in the DB.

mongodben (changes requested):
looks reasonable, but see comment about the approach and also add tests

chat-server/src/routes/conversations/createConversation.ts
```diff
@@ -39,7 +40,19 @@ export function makeCreateConversationRoute({
...
     const reqId = getRequestId(req);
     try {
-      const { ip } = req;
+      const {
+        headers: { origin: requestOrigin },
+        ip
+      } = req;
+
+      if (!requestOrigin) {
+        return sendErrorResponse({
```
mongodben, line 49: add test for this

chat-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -108,9 +94,17 @@ export function makeAddMessageToConversationRoute({
           User message: ${message}
           Stream: ${stream}
           IP: ${ip}
+          Origin: ${requestOrigin}
           ConversationId: ${conversationIdString}`,
       });
 
+      if (!requestOrigin) {
```
mongodben, line 101: add test for this

chat-server/src/routes/conversations/addMessageToConversation.ts
nlarew: I could switch this to middleware yes. Not sure exactly what you're referring to with an injected function / why we need it. I figure we can just treat this like `ip` and pass it to the service?

mongodben: > I figure we can just treat this like ip and pass it to the service?

we shouldn't require that we pass the IP either tbh.

the server core library should be as agnostic as possible about all the non-essential information that we pass to it, leaving that up to each implementation. 

let's get this change into main as it is now, and then we'll refactor in the feature branch for OSSification.

mongodben: > and then for persisting to the DB, i think we could make an injected function that takes the request object and adds arbitrary data from it (including origin) to the object in the DB.

thinking a bit more about this, and rather than injecting function, i think we could just include `ip`, `origin`, and whatever other request data that we want to persist to a custom property on the Express `Request` object, say `Request.customData`. 

Then, in the `POST /conversation` route, we put whatever is in the `Request.customData` object in the Conversation document in the DB, say in a `Conversation.customData` object.

mongodben (approved):

PR #239, opened 2023-11-13
Artifactory auth with token instead of username/password

nlarew:
## Changes

- Switch to `LG_ARTIFACTORY_TOKEN` from `LG_ARTIFACTORY_USERNAME` and `LG_ARTIFACTORY_PASSWORD`

**NOTE** - This only seems to work for `push` jobs. The `pr` job continues to 403, possibly because we have pull requests disabled for the secret in drone?


chat-server/trigger.me
mongodben, line 1: remove before merge pls

chat-ui/trigger.me
mongodben, line 1: remove before merge pls

ingest-service.dockerfile
mongodben, line 1: also same change to `chat-server.dockerfile` pls

mongodben (approved):
all seems reasonable to me. pls update the `chat-server.dockerfile` before merge as well.

PR #244, opened 2023-11-20
(EAI-138): Refactor origin handling, add custom conversation data logic, add conversations middleware

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-138

## Changes

-
-

## Notes

-

### DIFF
--- a/chat-server/src/routes/conversations/conversationsRouter.ts
+++ b/chat-server/src/routes/conversations/conversationsRouter.ts
@@ -1,10 +1,11 @@
-import { Request, Router } from "express";
+import { Request, Router, RequestHandler } from "express";
 import { rateLimit, Options as RateLimitOptions } from "express-rate-limit";
 import slowDown, { Options as SlowDownOptions } from "express-slow-down";
 import validateRequestSchema from "../../middleware/validateRequestSchema";
 import { ChatLlm } from "../../services/ChatLlm";
+import { Conversation } from "../../services/conversations";
 import { DataStreamer } from "../../services/dataStreamer";
-import { ConversationsService } from "../../services/conversations";
+import { ConversationCustomData, ConversationsService } from "../../services/conversations";
 import { RateMessageRequest, makeRateMessageRoute } from "./rateMessage";
 import {
   CreateConversationRequest,
@@ -48,6 +49,8 @@ export interface ConversationsRateLimitConfig {
   addMessageSlowDownConfig?: Partial<SlowDownOptions>;
 }
 
+export type AddCustomDataFunc = (request: Request) => Promise<ConversationCustomData>;
+
 /**
   Configuration for the /conversations/* routes.
  */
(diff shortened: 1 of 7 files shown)


chat-server/src/routes/conversations/conversationsRouter.ts
```diff
@@ -48,6 +54,37 @@ export interface ConversationsRateLimitConfig {
   addMessageSlowDownConfig?: Partial<SlowDownOptions>;
 }
 
+export type AddCustomDataFunc = (
```
nlarew, line 57: Should we have doc comment for this type too?

chat-server/src/routes/conversations/conversationsRouter.ts
```diff
@@ -77,6 +114,43 @@ export interface ConversationsRouterParams {
...
+  middleware?: ConversationsMiddleware[];
+
+  /**
+    Function that takes the request + response and returns any custom data you want to include
+    in the {@link Conversation} persisted to the database.
+    For example, you might want to store the user's email address with the conversation.
+    The custom data is persisted to the database with the Conversation in the
+    {@link Conversation.customData} field.
+   */
+  createConversationCustomData?: AddCustomDataFunc;
```
nlarew, line 144: I don't think it's a blocker, but it feels a bit weird to me to use these route-specific custom data middleware slots. Maybe in the future we can have a more generic "post route" middleware parameter that filters on `req` like your example code for `middleware`?

chat-server/src/routes/conversations/createConversation.ts
```diff
@@ -20,51 +20,67 @@ export const CreateConversationRequest = SomeExpressRequest.extend({
...
     try {
-      const { ip, origin: requestOrigin } = req;
-
-      if (!isValidIp(ip)) {
-        return sendErrorResponse({
+      logRequest({
+        reqId,
+        message: `Creating conversation`,
+      });
+      try {
```
nlarew, line 45: Can we un-nest this to use a single `try-catch`? This triple nested try block is smelly to me.

See how we do this in `addMessageToConversation`

chat-server/src/services/conversations.ts
```diff
@@ -132,10 +160,7 @@ export interface ConversationsService {
   }: RateMessageParams) => Promise<boolean>;
 }
 
-/**
- OSS_TODO: make these configurable from the entry point. though i think these messages are reasonable defaults
- */
-export const conversationConstants = {
+export const defaultConversationConstants = {
```
nlarew, line 163:
```suggestion
export const defaultConversationConstants: ConversationConstants = {
```

nitty but it'll make sure our default is aligned with the type def

nlarew (commented):
Looking pretty good! A few small comments before we ship

chat-server/src/routes/conversations/conversationsRouter.ts
mongodben: seems reasonable. will add

chat-server/src/routes/conversations/createConversation.ts
mongodben: implemented. lmk what you think

nlarew (approved):
LGTM

PR #248, opened 2023-11-29
(EAI-145) InputBar adds newlines on shift+enter

nlarew:
Jira: (EAI-145) [UI] InputBar does not add newlines on shift+enter

## Changes

- Now you can press {shift|ctrl}+enter to expand the input bar height (up to 5 lines)

inputbar


mongodben (approved):
note that i changed the title to better reflect what's in this PR

PR #263, opened 2023-12-07
Fix ingest cron

mongodben:
Jira: n/a

## Changes

- Small fixes to make ingest cron work again
-

## Notes

- It stopped working due to some changes in the repo

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -317,7 +317,6 @@ name: staging-build-ingest-mongodb-public
 trigger:
   branch:
     - main
-    - fix_ingest_cron
   event:
     - push
   paths:
@@ -378,7 +377,6 @@ trigger:
       - packages/mongodb-rag-core/**/*
   branch:
     - main
-    - fix_ingest_cron
 
 steps:
   # Deploys docker image associated with staging build that triggered promotion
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -377,6 +378,7 @@ trigger:
       - packages/mongodb-rag-core/**/*
   branch:
     - main
+    - fix_ingest_cron
```
mongodben, line 381:
```suggestion
```

remove before merge

packages/ingest-mongodb-public/package.json
```diff
@@ -15,6 +15,7 @@
     "watch": "tsc -b -w",
     "release": "release-it",
     "ingest:all": "ingest all --config ./build/config.js",
```
mongodben, line 17: interestingly, this `ingest:all` command works locally AND in a local docker image using the same ingest-service.dockerfile. but the `ingest` binary doesn't work in the k8s-deployed docker image for reasons i do not understand. will ask kanopy team if they can explain

mongodben: but in the meantime, we can merge this as-is i think

nlarew (approved):

PR #269, opened 2023-12-11
(EAI-174) [Core] Support arbitrary MongoDB queries in PageStore

nlarew:
Jira: (EAI-174) [Core] Support arbitrary MongoDB queries in PageStore

### DIFF
--- a/packages/mongodb-rag-core/src/MongoDbPageStore.ts
+++ b/packages/mongodb-rag-core/src/MongoDbPageStore.ts
@@ -4,7 +4,7 @@ import {
   MakeMongoDbDatabaseConnectionParams,
   makeMongoDbDatabaseConnection,
 } from "./MongoDbDatabaseConnection";
-import { LoadPagesArgs, LoadPagesQuery, PageStore, PersistedPage, parseLoadPagesArgs } from "./Page";
+import { LoadPagesArgs, PageStore, PersistedPage } from "./Page";
 import { Filter } from "mongodb";
 
 /**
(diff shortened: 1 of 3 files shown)


packages/mongodb-rag-core/src/Page.ts
```diff
@@ -54,26 +56,35 @@ export type PersistedPage = Page & {
   action: PageAction;
 };
 
+export type LoadPagesQuery = {
```
mongodben, line 59: why support both this and LoadPageArgs?

isn't LoadPagesArgs a subset of LoadPagesQuery?

mongodben: also maybe we could abstract this to a `Record<string, unknown>` here, and then in the makeMongoDbPageStore implementation use `Filter<PersistedPage>`. this way the Page type abstraction isn't tightly coupled to MongoDB.

packages/mongodb-rag-core/src/MongoDbPageStore.ts
```diff
@@ -20,16 +20,13 @@ export function makeMongoDbPageStore({
   });
   const pagesCollection = db.collection<PersistedPage>("pages");
   return {
+    queryType: "mongodb",
```
cbush, line 23: If it's possible, I would export a special type:

```ts
export type MongoDBPageStore = PageStore & DatabaseConnection & {
  queryType: "mongodb";
  loadPages(args?: LoadPagesArgs<Filter<PersistedPage>>): Promise<PersistedPage[]>;
};
```
Then users can do
```ts
if (store.queryType === "mongodb") {
  (store as MongoDbPageStore).loadPages({ ... }); // VS Code helps with 'filter' type
}
```
Return that type from makeMongoDbPageStore.

Note that I imply LoadPagesArgs could be `LoadPagesArgs<T = unknown>` where T is the type of `query`. SHOULD work? (Might not)

cbush (approved):

packages/mongodb-rag-core/src/MongoDbPageStore.ts
nlarew: Does work!

cbush (approved):

PR #272, opened 2023-12-12
Fix staging static site

mongodben:
Jira: n/a

## Changes

- fix issue where staging site wasn't building in staging b/c was .gitignoring a necessary config file
- make dockerfile build and deploy based on lerna so don't need to think about directories from docker

## Notes

- fix is already in staging


.drone.yml
```diff
@@ -66,6 +66,7 @@ name: staging-build-chat-server
 trigger:
   branch:
     - main
+    - fix_staging_static_site
```
mongodben, line 69:
```suggestion
```

remove before merge

.drone.yml
```diff
@@ -130,6 +131,7 @@ trigger:
       - packages/mongodb-chatbot-ui/**/*
   branch:
     - main
+    - fix_staging_static_site
```
mongodben, line 134:
```suggestion
```

remove before merge

package.json
```diff
@@ -16,7 +16,8 @@
     "lint": "lerna run lint",
     "release:chat-ui": "lerna run release --scope='mongodb-chatbot-ui' && git add ./package-lock.json && git commit --amend --no-edit",
     "scripts:analyzeMessages": "lerna run analyzeMessages --scope='scripts'",
-    "scripts:scrubMessages": "lerna run scrubMessages --scope='scripts'"
+    "scripts:scrubMessages": "lerna run scrubMessages --scope='scripts'",
+    "server:start": "lerna run start --scope='chatbot-server-mongodb-public'"
```
nlarew, line 20: nit: I'd format this like `start:server` e.g. how we do `npm run dev:server` and `npm run dev:ui` as subsets of `npm run dev`

nlarew (approved):

package.json
mongodben: fair. consistency would be good. scripts is the prefix on the above commands. will keep same here for now and make new ticket to update all

mongodben: https://jira.mongodb.org/browse/EAI-176

nlarew: To be clear I don't think we should switch the `scripts:<foo>` ones

PR #285, opened 2024-01-04
Artifact Generator

nlarew:
Jira: https://jira.mongodb.org/browse/EAI-129


packages/mongodb-rag-core/src/Page.ts
```diff
@@ -1,3 +1,5 @@
+import { Filter } from "mongodb";
+
```
mongodben, line 2: remove?

mongodben (approved):
LGTM. 1 required fix on the `Pages.ts` file, but otherwise just a few thoughts/suggestions.

not reviewing thoroughly since this was a prototype that we're not planning on actively maintaining or publishing (for now?)

PR #286, opened 2024-01-04
(EAI-184) [UI] Chatbot provider should allow multiple children

nlarew:
Jira: (EAI-184) [UI] Chatbot provider should allow multiple children

## Changes

- `Chatbot` now allows either one or multiple children (but not zero)

### DIFF
--- a/docs/docs/ui.md
+++ b/docs/docs/ui.md
@@ -58,14 +58,14 @@ The `mongodb-chatbot-ui` package exports the following components.
 
 The `<Chatbot />` component is effectively a React context provider that wraps your chatbot. It accepts the following props:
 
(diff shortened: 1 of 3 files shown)


mongodben (approved):

mongodben (changes requested):
sorry, un-LGTMing 😢. can you also update the docs as part of this change? 

2 locations:

- https://mongodb.github.io/chatbot/ui/#chatbot-root-component
- https://www.npmjs.com/package/mongodb-chatbot-ui

as an aside, we should perhaps consider removing the component reference from the README to keep a single source of truth.

mongodben (approved):

PR #292, opened 2024-01-10
(EAI-177): Implement FAQ Finder

cbush:
Jira: https://jira.mongodb.org/browse/EAI-177

## Changes

- Adds FAQ finding to the scripts package
- Using DBSCAN algorithm to cluster similar questions
- Script takes optional epsilon argument (otherwise defaults to 0.05). Epsilon is the max distance to form a cluster.
- Stores FAQ entries in a 'faq' collection in the database for further processing. Each question and response is stored even though many are similar.

Example FAQ entry document:
image

## Notes

The intention is to run this on cron maybe daily or every five days. Because each run stores the timestamp, we'll be able to look at trends in FAQ over time.

For verified answers, the next step is to script a transfer from the faq collection to a yaml document that can be filled out by a human, per the spec.


packages/scripts/src/clusterize.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+};
+
+export const distanceSquared = (a: number[], b: number[]) => {
+  assert(
+    a.length === b.length,
+    "Unexpected inconsistent dimensionality in distanceFunction"
+  );
+  // Pythagoras: the square of the hypotenuse is equal to the sum of the squares
+  // of the other sides
```
mongodben, line 67:
```suggestion
  // Pythagorean theorem: the square of the hypotenuse is equal to the sum of the squares
  // of the other sides
```

packages/scripts/src/findFaq.ts
mongodben, line 1: should this file have tests?

packages/scripts/src/analyzeMessages.ts
```diff
@@ -98,7 +98,7 @@ const analyzeMessages = async ({ db }: { db: Db }) => {
         embedding: 0,
       },
     },
-    { $sort: { conversationId: 1, createdAt: 1 } },
+    { $sort: { convoId: 1, createdAt: 1 } },
```
mongodben, line 101: why this change?

packages/scripts/src/barycenter.ts
mongodben, line 1: nit?: i'm very new to the clustering vectors stuff[^1], but per my brief googling, it seems that "barycenter" is generally more of a [physics/astronomy term](https://en.wikipedia.org/wiki/Barycenter_(astronomy)). I think [centroid](https://en.wikipedia.org/wiki/Centroid) could be a more accurate term to use here, as it's the same concept but in a purer mathematical sense.

perhaps `findCentroid` could be a better name for the file/function? what was the reason for choosing `barycenter`? 

[^1] but i did get an A- in linear algebra in college 💁).[^2]
[^2] and have been reading some DFW essays recently

packages/scripts/src/barycenter.ts
```diff
@@ -0,0 +1,20 @@
...
+export const barycenter = (vectors: number[][]) => {
+  assert(vectors.length > 0, "Cannot find barycenter of 0 vectors!");
+  assert(
+    vectors.find((vector) => vectors[0].length !== vector.length) === undefined,
+    "Vector dimensionality mismatch!"
+  );
+  return (
+    vectors
+      // Add all vectors together into one vector (add each dimension together)
+      .reduce((acc, cur) => cur.map((d, i) => (acc[i] ?? 0) + d), [])
```
mongodben, line 16: i don't think you need the `?? 0` in `acc[i] ?? 0` since you've already checked that all the vectors have the same length above.

packages/scripts/src/barycenter.ts
cbush: Centroid is good. I originally had "vectorMidpoint" but couldn't find "midpoint" as the term - barycenter came up (which is not strictly astronomical btw, but does imply 'center of gravity'. I think a figure of N equally weighted vertices will have the center of gravity == centroid).

packages/scripts/src/analyzeMessages.ts
cbush: Fixes a bug, note that the field is projected to "convoId" not "conversationId" above

packages/scripts/src/findFaq.ts
cbush: I don't think it really needs to since the meat of it is the other algorithms

packages/scripts/src/clusterize.ts
```diff
@@ -0,0 +1,69 @@
+import dbscan from "@cdxoo/dbscan";
+import { strict as assert } from "assert";
+
+// Based on https://www.npmjs.com/package/@cdxoo/dbscan
```
mongodben, line 4:
```suggestion
/**
  Based on https://www.npmjs.com/package/@cdxoo/dbscan
 */
```

packages/scripts/src/findCentroid.ts
```diff
@@ -0,0 +1,19 @@
...
+  Finds the centroid (mean) of a set of n-dimensional vectors.
+ */
+export const findCentroid = (vectors: number[][]) => {
+  assert(vectors.length > 0, "Cannot find centroid of 0 vectors!");
+  assert(
+    vectors.find((vector) => vectors[0].length !== vector.length) === undefined,
+    "Vector dimensionality mismatch!"
+  );
+  return (
+    vectors
```
mongodben, line 13: there are some libs to help w common linear algebra ops like this that'd make the code little more readable than these raw array operations. 

i worked w a Professional Service eng on something similar a few months back. can't find the lib he used, but asked him..will get back w it, and we can see if applicable here

cbush: It's done... and it's in the scripts.

Looked for n-dimensional vector libraries but didn't find much of quality. Obviously 2-4D is well-trodden ground.

packages/scripts/src/clusterize.ts
```diff
@@ -0,0 +1,71 @@
...
+};
+
+export const distanceSquared = (a: number[], b: number[]) => {
+  assert(
+    a.length === b.length,
+    "Unexpected inconsistent dimensionality in distanceFunction"
+  );
+  // Pythagorean theorem: the square of the hypotenuse is equal to the sum of
+  // the squares of the other sides
+  return a.reduce((acc, cur, i) => acc + Math.pow(b[i] - cur, 2), 0);
```
mongodben, line 70: ditto this comment https://github.com/mongodb/chatbot/pull/292/files#r1447931008

packages/scripts/src/clusterize.ts
```diff
@@ -0,0 +1,71 @@
...
+    distanceFunction(a, b) {
+      return distanceSquared(getVector(a), getVector(b));
+    },
+  });
+
+  return {
+    clusters: clusterIndexesArray.map((indexes) =>
+      indexes.map((index) => dataset[index])
+    ),
+    noise: noiseIndexes.map((index) => dataset[index]),
```
mongodben, line 59: what is `noise` in the context of clusters?

packages/scripts/src/clusterize.test.ts
mongodben, line 1: can you provide some explanatory comments to what's going on with this test? it's hard to follow

packages/scripts/src/clusterize.ts
cbush: Added docs above

packages/scripts/src/findFaq.ts
```diff
@@ -0,0 +1,139 @@
...
+
+import {
+  Conversation,
+  SomeMessage,
+  AssistantMessage,
+  FunctionMessage,
+} from "mongodb-chatbot-server";
+import { clusterize, DbscanOptions } from "./clusterize";
+import { findCentroid } from "./findCentroid";
+
```
mongodben, line 12: nit: i think you could make a new type for the intersection `AssistantMessage | FunctionMessage`:

```ts
type ResponseMessage = AssistantMessage | FunctionMessage;
```

packages/scripts/src/findFaq.ts
```diff
@@ -0,0 +1,139 @@
...
+export type QuestionAndResponses = {
+  embedding: number[];
+  question: string;
```
mongodben, line 15: might it be useful to keep the whole `UserMessage` object? 

that way you'd have the original `content`, `preprocessedContent` and `contentForLlm`.

packages/scripts/src/findFaq.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  const faqEntries = clusters
+    .map((cluster) => {
+      const embeddings = cluster.map(({ embedding }) => embedding);
+      return {
+        embedding: findCentroid(embeddings),
+        embeddings,
+        questions: cluster.map(({ question }) => question),
+        responses: cluster.map(({ responses }) => responses),
+        faqScore: cluster.length / (questions.length - noise.length),
```
mongodben, line 133: what is this metric?

packages/scripts/src/clusterize.test.ts
cbush: Added some comments

packages/scripts/src/findFaq.ts
cbush: This should the relative frequency of the question cluster

cbush: https://docs.google.com/document/d/1YQ8fcMneOPZGHbrowrKVEi-oHB82LnEtOlSuRJzsxRg/edit#heading=h.mbb4punhvs05

cbush: https://docs.google.com/document/d/1_uaiiVyNSQvTrVm9VxSZMfz590voNAf73PgYyckwOX0/edit#heading=h.vsfxxnuz34oo

packages/scripts/src/findFaq.ts
cbush: Def

packages/scripts/src/findCentroid.ts
mongodben: > It's done... and it's in the scripts.

what do you mean?

mongodben (changes requested):
looks great, just a few things throughout

packages/scripts/src/findCentroid.ts
cbush: 1. Let's not worry *too* much about the legibility/maintainability of the scripts package
2. I already implemented and tested the minimal vector math I need, so I'm less enthused to swap it for some library. As I mentioned, I looked for libraries but good n-dimensional ones were not jumping out to me

packages/scripts/src/findFaq.ts
mongodben: ah ok. why `questions.length - noise.length` instead of `questions.length`? why remove the cluster-less "noise" questions from the score?

packages/scripts/src/clusterize.ts
```diff
@@ -0,0 +1,89 @@
...
+    Maximum distance between datapoints. Determine if a datapoint is in a
+    cluster or not. Default is 1.0.
+   */
+  epsilon: number;
+
+  /**
+    Custom function to compare calculated distance and epsilon. Default is
+    (dist, e) => (dist < e)
+   */
+  epsilonCompare: (distance: number, epsilon: number) => boolean;
```
nlarew, line 18: What's the use case for a custom epsilon comparison? Isn't a core point neighborhood defined as d < e by the algorithm?

packages/scripts/src/findFaq.ts
cbush: Hmmm I had original just `/ questions.length` but then I liked the numbers I was getting without the noise more. The noise questions don't get saved in the faq collection, so it seemed sensible not to represent them in the faq score. That way you can take every entry from a findFaq run and sum all the faqScores to get 1. But it's completely arbitrary, I'm open to putting it back if you feel otherwise.

packages/scripts/src/clusterize.ts
cbush: This just comes from the dbscan library and I suspect it's so that you can do stuff like compare an 'actual length' epsilon with a 'magnitude' (e.g. `e < sqrt(distance)`). Note I don't bother sqrting the distances and just expect e to fit accordingly. Maybe you can also do `d > e` to make anti-clusters. I dunno! I didn't see a reason NOT to expose it so I just kept it in.

packages/scripts/src/findCentroid.ts
mongodben: ok i got the library, it's TensorFlow.js - https://js.tensorflow.org/api/4.15.0/ 

it actually uses hardware acceleration, so theoretically runs faster/more efficiently than standard array ops in addition to having a clearer API for these operations. 

for example, here's the ChatGPT-assisted version of findCentroid with TF.js:

```ts
export async function findCentroid(vectors: number[][]) {
  const meanTensor = tf.tensor(vectors).mean(0);
  return await centroidTensor.array();
}
```

packages/scripts/src/clusterize.ts
mongodben: and again ChatGPT generated answer w/ tensorflow:

```ts
export const distanceSquared = async (a: number[], b: number[]) => {
  // Convert arrays to tensors if they are not already
  const tensorA = tf.tensor(a);
  const tensorB = tf.tensor(b);

  // Check dimensions
  if (tensorA.shape[0] !== tensorB.shape[0]) {
    throw new Error("Unexpected inconsistent dimensionality in distanceFunction");
  }

  // Calculate squared distance
  const squaredDistanceTensor = tensorA.sub(tensorB).square().sum();

  // Get the result as a number
  const squaredDistance = await squaredDistanceTensor.array();
  
  // Since the result is a scalar, we can return the first element of the array
  return squaredDistance[0];
};
```

packages/scripts/src/clusterize.ts
```diff
@@ -0,0 +1,89 @@
...
+  const options: DbscanOptions = {
+    ...defaultDbscanOptions,
+    ...(optionsIn ?? {}),
+  };
+
+  const { clusters: clusterIndexesArray, noise: noiseIndexes } = dbscan({
+    ...options,
+    dataset,
+    distanceFunction(a, b) {
+      return distanceSquared(getVector(a), getVector(b));
```
nlarew, line 69: This is a Euclidean distance metric right?

My (limited) understanding is that in a many-dimensional space the Euclidean metric distance between _any_ two given vectors is ~1. i.e. the usefulness of the Euclidean metric is inversely proportional to number of dimensions

Should we use cosine similarity instead?

packages/scripts/src/findCentroid.ts
cbush: How strongly do you feel about this? Will Kanopy or wherever this runs noticeably benefit from (let alone be able to take advantage of) hardware acceleration? I don't really feel strongly at all, except that my implementation is done, works, is tested, is pretty straightforward tbh, is fit for purpose, is fast enough on my computer, etc, so swapping it for another library now smacks of premature optimization and goldplating...

packages/scripts/src/clusterize.ts
cbush: Maybe? Check the results in the dev database, seems to do a pretty good job at distinguishing different questions and grouping similar questions with a reasonable epsilon value. Open to suggestions though.

packages/scripts/src/clusterize.ts
cbush: Pretty cool, I wish I had thought of tensorflow before

packages/scripts/src/findFaq.ts
mongodben: the comparative frequency of clusters won't differ regardless of if we include `- noise.length` since noise.length is constant for all clusters in a findFaq() run, so i don't think it particularly matter which route we go. but with that being said, i personally prefer `cluster.length / questions.length` since it's simpler, and the idea of numQsInCluster/TotNumQs logically seems more straightforward

packages/scripts/src/clusterize.ts
nlarew: Not a blocker here, especially if our results are satisfying. Maybe something to experiment with in future!

packages/scripts/src/findFaq.ts
cbush: Yeah, I think it might be nice to have faqScore include the invisible NON-'faq' questions... so I'll put it back the way it was!

packages/scripts/src/findCentroid.ts
mongodben: i also don't feel strongly here. hardware acceleration aspect might matter when number of queries we're processing increases dramatically, but that can be a future problem. 

re kanopy taking advantage of the hardware acceleration, i believe it would optimize CPU use with tensorflow c++ bindings but not GPU acceleration

personally, i find using tensorflow clearer to read and advocate it for that reason if nothing else. 

and less direct to this PR, i think it's valuable for our team to start familiarizing ourselves with ML libraries and techniques like tensorflow as we dive deeper into the AI/ML space.

not going to block the PR for this.

mongodben (approved):
nice work here!

cbush:
I haven't pushed upstreammmm

packages/scripts/src/findCentroid.ts
cbush: Just for the sake of completeness, I tried your code out, but it doesn't work. I wanted to avoid going down a rabbit hole with a new library, but the docs say mean() is a reduction function that collapses a vector into a scalar. I tried with the keepDims parameter, but it doesn't work. I looked for another function that would find the centroid, but couldn't find one. This is why I dislike the attitude that one should never do their own math when a library exists. There is such a thing as overkill.

PR #305, opened 2024-01-23
Fix cron names

cbush:
- Apparently they must be lowercase


packages/scripts/environments/production.yml
```diff
@@ -42,7 +42,7 @@ cronJobs:
         cpu: 500m
         memory: 5Gi
 
-  - name: findFaq
+  - name: findfaq
```
mongodben, line 45: nit: maybe snake_case or kebab-case here?

mongodben (approved):
lgtm w small comment on name change

PR #307, opened 2024-01-23
(EAI-193) Qualitative Feedback

nlarew:
Jira: (EAI-193) Qualitative Feedback

This allows users to leave an optional text comment to qualify their thumbs up/down ratings.

## Changes

- Adds `userComment` field to assistant messages in the database
- Adds a new server endpoint `/conversations/{conversationId}/messages/{messageId}/comment`
- Adds UI to support this interaction

## Notes

- Lots of hacky fixes in the UI due to some outstanding bugs in the MessageFeedback component. See the related issues linked from EAI-193
- Also includes some incidental CSS changes to have more consistent spacing in markdown lists

https://github.com/mongodb/chatbot/assets/15657698/23e3ebdd-f85f-477b-9921-c755c2e9e231

Screenshot 2024-01-23 at 4 00 21 PM


packages/mongodb-chatbot-server/src/services/ConversationsService.ts
```diff
@@ -211,6 +224,15 @@ export interface ConversationsService {
     messageId,
     rating,
   }: RateMessageParams) => Promise<boolean>;
+
+  /**
+    Add a user comment to a {@link Message} that already has a rating.
```
mongodben, line 229: nit: msg doesn't need a rating to add a comment

 ```suggestion
    Add a user comment to a {@link Message}.
```

File: packages/mongodb-chatbot-server/src/services/mongodbConversations.test.ts [hunk 2]
```diff
@@ -221,4 +221,126 @@ describe("Conversations Service", () => {
...
+        references: [],
+      },
+    });
+
+    const rateResult = await conversationsService.rateMessage({
+      conversationId,
+      messageId: assistantMessage.id,
+      rating: true,
+    });
+    expect(rateResult).toBe(true);
```
mongodben, line 250: why is this in the test?

packages/mongodb-chatbot-server/src/services/mongodbConversations.ts
```diff
@@ -126,6 +127,39 @@ export function makeMongoDbConversationsService(
...
+      comment,
+    }: CommentMessageParams) {
+      const updateResult = await conversationsCollection.updateOne(
+        {
+          _id: conversationId,
+          messages: {
+            $elemMatch: {
+              id: messageId,
+              role: "assistant",
+              rating: { $exists: true },
```
mongodben, line 143: as noted here https://github.com/mongodb/chatbot/pull/307/files#r1464163336, my inclination is to not require rating to exist at the ConversationService level

packages/mongodb-chatbot-server/src/services/mongodbConversations.ts
```diff
@@ -126,6 +127,39 @@ export function makeMongoDbConversationsService(
...
+              userComment: { $exists: false },
```
mongodben, line 144: think it could make sense to not have this as well so that a message can be updated even if it already exists

packages/mongodb-chatbot-server/src/routes/conversations/commentMessage.ts
```diff
@@ -0,0 +1,134 @@
...
+  z.object({
+    headers: z.object({
+      "req-id": z.string(),
+    }),
+    params: z.object({
+      conversationId: z.string(),
+      messageId: z.string(),
+    }),
+    body: z.object({
+      comment: z.string().min(1, { message: "Comment cannot be empty" }),
```
mongodben, line 28: i think there should also be some max length for the comment.

not sure if it should be here at the schema level or the route level

mongodben (commented):
please also add docs for this endpoint. i think that'd be adding to:

1. openapi spec https://mongodb.github.io/chatbot/server/openapi
2. list of endpoints https://mongodb.github.io/chatbot/server/conversations#endpoints
   - alternately, to reduce docs duplication and avoid drift, we might want to have this section just point to the openapi spec w/o listing what the available endpoints are

packages/mongodb-chatbot-ui/src/Message.tsx
```diff
@@ -130,41 +162,81 @@ export const Message = ({
...
+        messageRatingProps={{
+          className: styles.message_rating,
+          description: "How was the response?",
+          onChange: async (e) => {
+            const value = e.target.value as MessageRatingProps["value"];
+            if (!value) {
+              return;
+            }
+            if (ratingCommentStatus === "submitted") {
+              // Once a user has submitted a comment for their rating we don't want them to be able to change their rating
```
mongodben, line 203: why not let them update?

packages/mongodb-chatbot-ui/src/Message.tsx
```diff
@@ -260,3 +332,110 @@ export const MessagePrompts = ({
...
+      className={css`
+        & > div + div {
+          margin-top: 0.5rem;
+        }
+      `}
+    >
+      <MessageRating value={value} {...messageRatingProps} />
+      {hasRating && status !== "abandoned" ? (
+        <>
+          <InlineMessageFeedback
```
mongodben, line 378: can you add a max input length FE validation to the text area? presumably same as what's on the server. 

if server is configurable, then i guess that length should be here too.

packages/mongodb-chatbot-ui/src/Message.tsx
nlarew: At a basic level because that's how it's designed https://www.figma.com/file/L41rvEwQmJJ8VzO4hCPc7Y/LeafyGreen-Chat-Library?type=design&node-id=1-493&mode=design&t=1vkDz6xoVoPuaIIH-0

The motivation being:

1. User rates the message NEGATIVE
2. User submits a comment like "This was bad because it recommended PHP when I asked for Python"
3. User re-rates the message POSITIVE
4. Now there's a mismatch between the rating and the (uneditable) comment

packages/mongodb-chatbot-server/src/services/ConversationsService.ts
nlarew: Hmm yeah that's reasonable I can look to adjust it

packages/mongodb-chatbot-ui/src/Message.tsx
mongodben: ok, my instinct would be to have it all be editable but i defer to designs

mongodben (changes requested):
generally looks good, just some smaller stuff on the server + docs. i can also manually test the UI once the server stuff is implemented.

packages/mongodb-chatbot-server/src/routes/conversations/commentMessage.ts
nlarew: Added to the route config

packages/mongodb-chatbot-ui/src/Message.tsx
nlarew: It's a little hacky since this isn't baked into the LG component but it works!

docs/docs/server/openapi.yaml
```diff
@@ -96,6 +96,28 @@ paths:
         500:
           $ref: "#/components/responses/InternalServerError"
 
+  /conversations/{conversationId}/messages/{messageId}/comment:
+    post:
+      operationId: commentMessage
+      summary: Add a comment to an assistant message that clarifies a thumbs up/down rating.
```
mongodben, line 102:
```suggestion
      summary: Add comment to an assistant message
      description: Add a comment to an assistant message that clarifies a thumbs up/down rating.
```

packages/mongodb-chatbot-server/src/routes/conversations/commentMessage.ts
```diff
@@ -0,0 +1,158 @@
...
+      if (!conversationInDb) {
+        return sendErrorResponse({
+          reqId,
+          res,
+          httpStatus: 404,
+          errorMessage: "Conversation not found",
+        });
+      }
+
+      const existingMessage = conversationInDb.messages.find((message) =>
```
mongodben, line 92: since user's will probably be commenting on msgs near the end of the comment array, consider using [`Array.prototype.findLast()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast) instead of `Array.prototype.find()`.

packages/mongodb-chatbot-server/src/services/ConversationsService.ts
```diff
@@ -32,6 +32,14 @@ export type AssistantMessage = MessageBase & {
    */
   rating?: boolean;
 
+  /**
+    An additional text comment provided by the user to clarify their
+    `rating` for the message. If `rating` is undefined, this should be
+    too. Note that only messages with `role: "assistant"` can be rated
+    and commented on.
```
mongodben, line 39:
```suggestion
    too.
```

nit: not nec here b/c this is only on assistant msg type

mongodben (approved):
LGTM. few small comments about docs + 1 optimization

PR #308, opened 2024-01-24
(EAI-206): Add authorized user to conversation custom data if env != prod && authorized user cookie exists

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-206

## Changes

- Add ability to get authorized user from the cookie that CorpSecure sets in lower level envs. 
- Explicitly add `requireRequestOrigin()` and `requireValidIpAddress()` middleware and logic to add these to the conversation custom data b/c the new functionality to get the auth user overrides the defaults which used these. 

## Notes

- To use this functionality, you **must** access the site behind CorpSecure, such as our staging URL http://chat-server.docs.staging.corp.mongodb.com/
  - However, the staging site that sits behind CDN + firewall will not work for this functionality (https://knowledge.staging.corp.mongodb.com/). This site isn't behind CorpSecure, just IP-restricted firewall.

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/config.ts
+++ b/packages/chatbot-server-mongodb-public/src/config.ts
@@ -24,10 +24,14 @@ import {
   MakeUserMessageFuncParams,
   UserMessage,
   makeFilterNPreviousMessages,
+  requireRequestOrigin,
+  requireValidIpAddress,
+  ConversationCustomData,
 } from "mongodb-chatbot-server";
 import { stripIndents } from "common-tags";
 import { makePreprocessMongoDbUserQuery } from "./processors/makePreprocessMongoDbUserQuery";
 import { AzureKeyCredential, OpenAIClient } from "@azure/openai";
+import cookieParser from "cookie-parser";
 
 export const {
   MONGODB_CONNECTION_URI,
@@ -202,11 +206,30 @@ export const conversations = makeMongoDbConversationsService(
   systemPrompt
 );
 
+const isProduction = process.env.NODE_ENV === "production";
 export const config: AppConfig = {
   conversationsRouterConfig: {
     dataStreamer,
     llm,
     conversations,
+    middleware: [
+      requireRequestOrigin(),
+      requireValidIpAddress(),
+      cookieParser(),
+    ],
+    createConversationCustomData: async (req, res) => {
+      const customData: ConversationCustomData = {};
+      if (!isProduction && req.cookies.auth_user) {
+        customData.authUser = req.cookies.auth_user;
+      }
+      if (req.ip) {
+        customData.ip = req.ip;
+      }
+      if (res.locals.customData.origin) {
+        customData.origin = res.locals.customData.origin;
+      }
+      return customData;
+    },
     generateUserPrompt,
     maxUserMessagesInConversation: 50,
     filterPreviousMessages: makeFilterNPreviousMessages(12),
@@ -215,5 +238,5 @@ export const config: AppConfig = {
   corsOptions: {
     origin: allowedOrigins,
   },
-  serveStaticSite: process.env.NODE_ENV !== "production",
+  serveStaticSite: !isProduction,
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -13181,6 +13181,15 @@
         "@types/express": "*"
       }
     },
+    "node_modules/@types/cookie-parser": {
+      "version": "1.4.6",
+      "resolved": "https://registry.npmjs.org/@types/cookie-parser/-/cookie-parser-1.4.6.tgz",
+      "integrity": "sha512-KoooCrD56qlLskXPLGUiJxOMnv5l/8m7cQD2OxJ73NPMhuSz9PmvwRD6EpjDyKBVrdJDdQ4bQK7JFNHnNmax0w==",
+      "dev": true,
+      "dependencies": {
+        "@types/express": "*"
+      }
+    },
     "node_modules/@types/cookiejar": {
       "version": "2.1.5",
       "resolved": "https://registry.npmjs.org/@types/cookiejar/-/cookiejar-2.1.5.tgz",
@@ -17074,6 +17083,26 @@
       "version": "0.5.0",
       "resolved": "https://registry.npmjs.org/cookie/-/cookie-0.5.0.tgz",
       "integrity": "sha512-YZ3GUyn/o8gfKJlnlX7g7xq4gyO6OSuhGPKaaGssGB2qgDUS0gPgtTvoyZLTt9Ab6dC4hfc9dV5arkvc/OCmrw==",
+      "engines": {
+        "node": ">= 0.6"
+      }
+    },
+    "node_modules/cookie-parser": {
+      "version": "1.4.6",
+      "resolved": "https://registry.npmjs.org/cookie-parser/-/cookie-parser-1.4.6.tgz",
+      "integrity": "sha512-z3IzaNjdwUC2olLIB5/ITd0/setiaFMLYiZJle7xg5Fe9KWAceil7xszYfHHBtDFYLSgJduS2Ty0P1uJdPDJeA==",
+      "dependencies": {
+        "cookie": "0.4.1",
+        "cookie-signature": "1.0.6"
+      },
+      "engines": {
+        "node": ">= 0.8.0"
+      }
+    },
+    "node_modules/cookie-parser/node_modules/cookie": {
+      "version": "0.4.1",
+      "resolved": "https://registry.npmjs.org/cookie/-/cookie-0.4.1.tgz",
+      "integrity": "sha512-ZwrFkGJxUR3EIoXtO+yVE69Eb7KlixbaeAWfBQB9vVsNn/o+Yw69gBWSSDK825hQNdN+wF8zELf3dFNl/kxkUA==",
       "engines": {
         "node": ">= 0.6"
       }
@@ -38301,6 +38330,7 @@
       "dependencies": {
         "@azure/openai": "^1.0.0-beta.5",
         "common-tags": "^1.8.2",
+        "cookie-parser": "^1.4.6",
         "cors": "^2.8.5",
         "dotenv": "^16.0.3",
         "express": "^4.18.2",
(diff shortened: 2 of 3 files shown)


packages/mongodb-chatbot-ui/src/Chatbot.tsx
```diff
@@ -17,6 +18,7 @@ export type ChatbotProps = {
   shouldStream?: boolean;
   tck?: string;
   user?: User;
+  fetchOptions?: ConversationFetchOptions;
```
mongodben, line 21: @nlarew react-y question: rather than passing object of options here, wondering if you think there could be a better way of doing this for if the stuff in that object changes. 

like if the user is having this be an object regulated by React state (say if auth token changes), right now i think that'd cause a whole chatbot re-render, which i don't think is what we'd want here.

nlarew: re: memoization, you have to be careful passing objects as props because they can be recreated every render and lose their identity.

For example consider this code:

```
// A. object props defined inline
<Chatbot fetchOptions={{ headers: { "X-My-Api-Key": "abcdefg" } }}>
  {children}
</Chatbot>

// B. static object
// Define customFetchOptions outside of the component
const customFetchOptions = { headers: { "X-My-Api-Key": "abcdefg" } };
// ... later on in the component render
<Chatbot fetchOptions={customFetchOptions}>
  {children}
</Chatbot>
```

Version A will naively cause a full tree rerender every time because the value of `fetchOptions` is a new object (with the same keys & values) every render. We would need to add additional equality checks / memoization inside our component tree.

Version B will share the identity of `customFetchOptions` between rerenders.

packages/mongodb-chatbot-ui/src/services/conversations.ts
mongodben, line 1: note to self: test this

packages/mongodb-chatbot-ui/src/Chatbot.tsx
mongodben: per convo, let's keep as is w/ option B.

packages/mongodb-chatbot-server/src/services/dataStreamer.ts
```diff
@@ -24,7 +24,6 @@ function makeServerSentEventDispatcher<
       res.writeHead(200, {
         "Cache-Control": "no-cache",
         "Content-Type": "text/event-stream",
-        "Access-Control-Allow-Origin": "*",
```
mongodben, line 27: needed to remove this to have CORS work. if i didn't remove this, then the browser was showing this CORS err:

```
Access to fetch at 'http://localhost:3000/api/v1/conversations/65bacd2f27c07aaa82302c05/messages?stream=true' 
from origin 'http://localhost:5173' has been blocked by CORS policy: 
The value of the 'Access-Control-Allow-Origin' header in the response 
must not be the wildcard '*' when the request's credentials mode is 'include'.
```

packages/chatbot-server-mongodb-public/src/config.ts
```diff
@@ -129,18 +136,47 @@ export const conversations = makeMongoDbConversationsService(
...
+      ? createCustomConversationDataWithIpAuthUserAndOrigin
+      : undefined,
     generateUserPrompt,
     maxUserMessagesInConversation: 50,
     maxUserCommentLength: 500,
   },
   maxRequestTimeoutMs: 30000,
   corsOptions: {
     origin: allowedOrigins,
+    credentials: !isProduction,
```
nlarew, line 179: AFAIK this means that we allow requests to include cookies in non-prod environments.

Since CORS is kind of tricky can we add a comment here explaining this?

Also why are we allowing cookies in dev/staging but not prod?

packages/mongodb-chatbot-server/src/services/dataStreamer.ts
nlarew: As is we're only including credentials in non-prod environments. Will this still work in prod envs where we aren't allowing cookies/credentials?

packages/mongodb-chatbot-ui/src/App.tsx
```diff
@@ -38,6 +38,7 @@ function App() {
           name="MongoDB AI"
           shouldStream={shouldStream}
           darkMode={darkMode}
+          fetchOptions={{ credentials: "include" }}
```
nlarew, line 41: Should we add to DevCenterChatbot's wrapper below this too?

packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -62,12 +62,25 @@ export class TimeoutError<Data extends object = object> extends Error {
...
 
+/**
+  Options to include with every fetch request made by the ConversationService.
+  This can be used to set headers, etc.
+ */
+export type ConversationFetchOptions = Omit<
+  RequestInit,
+  "body" | "method" | "headers" | "signal"
+> & {
+  headers?: Record<string, string>;
```
nlarew, line 73: Should we accept a standard [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object here?

I'm personally partial to using web standards when possible. In this case it can guarantee that headers are well-formed and in general is less for us to document and maintain.

packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -102,8 +122,10 @@ export class ConversationService {
   async createConversation(): Promise<Required<ConversationState>> {
     const path = `/conversations`;
     const resp = await fetch(this.getUrl(path), {
+      ...this.fetchOptions,
       method: "POST",
       headers: {
+        ...this.fetchOptions.headers,
```
nlarew, line 128: If we did accept a `Headers` object, we could use a lib like https://github.com/whitecrownclown/merge-headers for this step

packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -134,10 +156,12 @@ export class ConversationService {
   }): Promise<MessageData> {
     const path = `/conversations/${conversationId}/messages`;
     const resp = await fetch(this.getUrl(path), {
+      ...this.fetchOptions,
       method: "POST",
       headers: {
         "Content-Type": "application/json",
         [CUSTOM_REQUEST_ORIGIN_HEADER]: getCustomRequestOrigin() ?? "",
+        ...this.fetchOptions.headers,
```
nlarew, line 164: For `createConversation` we spread `fetchOptions.headers` _under_ our `Content-Type` and custom request origin headers (i.e. we overwrite the custom headers with our own). Here in `addMessage` we spread _over_ our headers which is the opposite (i.e. we let users overwrite our headers with their own).

Should we be consistent about this across methods?

packages/mongodb-chatbot-ui/src/services/conversations.ts
mongodben: there was some reason why i didn't..don't recall what it was..probably not a great reason. will reapproach

packages/mongodb-chatbot-server/src/services/dataStreamer.ts
mongodben: it was when i played around w/ it in local dev, but maybe a solution is to let the headers here be configurable when creating the data streamer. that way we can keep as is for prod and mutate as we wish for staging

packages/mongodb-chatbot-ui/src/App.tsx
mongodben: yeah good catch

packages/chatbot-server-mongodb-public/src/config.ts
mongodben: > Also why are we allowing cookies in dev/staging but not prod?

allow in dev/staging b/c we use to get user from cookie.

whereas we don't do this in prod b/c we don't need to. realistically, i don't think there's any threat vector here b/c the prod server isn't touching cookies. my initial thought was only allow for low lvls b/c of principle of least privilege, but it probably doesn't meaningfully matter here.

packages/mongodb-chatbot-ui/src/services/conversations.ts
mongodben: took a bit of typescript-foo to use the headers object but got it working

packages/chatbot-server-mongodb-public/src/config.ts
mongodben: adding comment and have it always accept credentials b/c upon research and analysis, there's no threat vector to us by allowing browser to send credentials.

packages/mongodb-chatbot-server/src/services/dataStreamer.ts
mongodben: given we're now allowing credentials (including cookies) per https://github.com/mongodb/chatbot/pull/308#discussion_r1492822479, this shouldn't be a concern

mongodben: > it was when i played around w/ it in local dev, but maybe a solution is to let the headers here be configurable when creating the data streamer. that way we can keep as is for prod and mutate as we wish for staging

i'd rather not add this b/c i don't think it's necessary and i'd rather not add a feature for the sake of adding the feature.

packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -62,12 +62,25 @@ export class TimeoutError<Data extends object = object> extends Error {
...
+};
+
 export type ConversationServiceConfig = {
   serverUrl: string;
+  fetchOptions?: ConversationFetchOptions;
 };
 
 export class ConversationService {
   private serverUrl: string;
+  private fetchOptions?: ConversationFetchOptions;
```
nlarew, line 83: We define `this.fetchOptions` in the class constructor so it should always be defined right?

Maybe an opportunity for cleanup in other spots in the class where we currently assume `fetchOptions` may be undefined and use nullish coalescing (i.e. `??`)

packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -222,12 +254,11 @@ export class ConversationService {
     };
 
     await fetchEventSource(this.getUrl(path, { stream: "true" }), {
+      ...this.fetchOptions,
+      // Need to convert Headers to plain object for fetchEventSource
+      headers: Object.fromEntries(this.fetchOptions?.headers ?? new Headers()),
```
nlarew, line 259: Shouldn't `this.fetchOptions` and `this.fetchOptions.headers` always be defined since we assign it in the constructor?

nlarew (approved):
Couple of small changes in the frontend code but then I think this should be good to go! Tests seems to indicate the the overrides work well. Let's double check in staging after the merge to make sure and then we can cut new releases.

PR #309, opened 2024-01-24
(EAI-154): Fix GitHubLoader-based sources

cbush:
Jira: https://jira.mongodb.org/browse/EAI-154

## Changes

- Fixes the missing leading slash in various GitHubDataSource-based sources' file filters and url transformers
- Adds tests
- Removes some unneeded awaits
- Adds note to docs about the leading slash

## Notes

I think the leading slash was not there until the implementation was swapped out for Git loader...


packages/ingest-mongodb-public/src/sources/pyMongo.test.ts
mongodben, line 1: we should add some basic make-sure-it-works test like this to all the data source implementations that aren't tested in mongodb-rag-ingest (not in this PR)

packages/ingest-mongodb-public/src/sources/index.test.ts
mongodben, line 1: thanks for adding these tests.

it feels odd to me that they're in an `index.test.ts` file. consider creating separate files for each of these data sources + accompanying test files, as you did for pymongo.

mongodben (approved):
LGTM. see note about moving some stuff around

PR #314, opened 2024-01-31
(EAI-195): Get a conversation by ID route

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-195

## Changes

- Add route to get a conversation by it's ID in the framework, `GET /conversations/:conversationId`
- Document the new route
- Block that route in our implementation since we haven't thought about security on top of ability to get conversation data.

## Notes

- Not adding ability to get custom data with the route for now even though we discussed in the ticket. We can add later if needed.

### DIFF
--- a/docs/docs/server/conversations.md
+++ b/docs/docs/server/conversations.md
@@ -19,6 +19,14 @@ The server has the following endpoints:
 
 For more information on these endpoints, refer to the [API Specification](openapi).
 
+## Security
+
+If you do not add custom middleware that handles authentication/authorization,
+a malicious actor who gets a conversation ID can modify the conversation.
+You can also add custom middleware to block requests to endpoints if you do not need them in your app.
+
+For more information about securing your server with custom middleware, refer to the [custom middleware request validation](custom-logic.md#request-validation) documentation.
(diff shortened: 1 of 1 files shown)


nlarew:
This is https://jira.mongodb.org/browse/EAI-195?

docs/docs/server/conversations.md
```diff
@@ -11,6 +11,7 @@ The server has the following endpoints:
   conversation before you can send messages to the chatbot.
 - `POST /conversations/:conversationId/messages` sends a message to the chatbot,
   and gets a response back.
+- `GET /conversations/:conversationId` gets a conversation in its current state.
```
nlarew, line 14: Do we want to note the security implications here or elsewhere?

i.e. in a naïve implementation, if a malicious actor got a hold of your conversation id they could read your entire conversation history so you should probably add some middleware to enforce authn/authz.

nlarew (approved):

docs/docs/server/conversations.md
mongodben: good point. will add documentation

PR #316, opened 2024-01-31
(EAI-178): Implement Verified Answers Importer

cbush:
Jira: https://jira.mongodb.org/browse/EAI-178

## Changes

- Adds importVerifiedAnswers to scripts

## Notes

- Correctly creates, updates, and deletes according to yaml
- Validates yaml using zod

## TODO (this PR)

- [x] Needs unit tests
- [x] Might want to make this its own package/cli before merging


packages/scripts/assets/verifiedAnswers/mongodb-verified-answers.yaml
```diff
@@ -0,0 +1,16 @@
+- questions:
+    - How do I verify answers?
+  answer: To verify an answer...
+  author_email: someone@example.com
+  references:
+    - uri: http://example.com
+      title: Example
+  hidden: true
```
nlarew, line 8: Are we concerned about verified answers becoming outdated? Should we also have some sort of field that marks the answers as valid/verified until some condition is met? e.g.

- `valid_until: new Date("June 30 2024")`
- `valid_for_versions: ["6.2", "7.0"]`

I think the date based approach would be easiest to implement.

With that in place we could set up a cron that emails the `author_email` (and maybe us too) asking them to review their answers & either delete them or "revalidate" for another period of time.

packages/scripts/src/verifiedAnswers/parseVerifiedAnswersYaml.ts
```diff
@@ -0,0 +1,31 @@
+import { parse } from "yaml";
+import { z } from "zod";
+
+export const referenceSchema = z.object({
+  uri: z.string(),
+  title: z.optional(z.string()),
+});
+
+export type Reference = z.infer<typeof referenceSchema>;
```
nlarew, line 9: Minor - we could base these on the `rag-core` definition. https://github.com/mongodb/chatbot/blob/main/packages/mongodb-rag-core/src/services/conversations.ts

If we want title to be optional I think you can extend & override like this:

```ts
const existingSchema = z.object({
  fieldName: z.string()
});

const newSchema = existingSchema.extend({
  fieldName: z.string().optional()
});
```

packages/scripts/assets/verifiedAnswers/mongodb-verified-answers.yaml
cbush: Captured: https://jira.mongodb.org/browse/EAI-224

packages/mongodb-verified-answers/.env.example
```diff
@@ -0,0 +1,18 @@
...
+MONGODB_CONNECTION_URI=<connection_uri>
+MONGODB_DATABASE_NAME=<docs-chatbot-ENV>
+VECTOR_SEARCH_INDEX_NAME=<index_name>
+NODE_ENV=development
+# Connection string for DevCenter
+DEVCENTER_CONNECTION_URI="mongodb+srv://<username>:<password>@devhub-cluster.sewho.mongodb.net/?retryWrites=true&w=majority"
+# For connecting to MongoDB University Data API
+UNIVERSITY_DATA_API_KEY=abc123
+# For use with GitHubDataSource. The token only needs read access to the repo.
+GITHUB_ACCESS_TOKEN=ghp_TOKENTOKENTOKEN
```
nlarew, line 18:
```suggestion
OPENAI_ENDPOINT=https://<resource_name>.openai.azure.com/
OPENAI_API_KEY=<api_key>
OPENAI_EMBEDDING_DEPLOYMENT=<deployment_name>
OPENAI_EMBEDDING_MODEL=text-embedding-ada-002
OPENAI_EMBEDDING_MODEL_VERSION=2023-03-15-preview
MONGODB_CONNECTION_URI=<connection_uri>
MONGODB_DATABASE_NAME=<docs-chatbot-ENV>
```

I think we don't need a good chunk of these. Might want to double check that this list is correct - I based it off of the single call to `assertEnvVars`.

packages/mongodb-verified-answers/README.md
```diff
@@ -0,0 +1,3 @@
+# MongoDB Verified Answer CLI
+
+The Verified Answer CLI imports answers specified in a yaml file into the database.
```
nlarew, line 3: Probably some more info we could put in the README (install, usage example, etc) but fine to leave minimal for now IMO

nlarew (approved):
LGTM - let's update the example `.env` file before merge

packages/mongodb-verified-answers/src/commands/import.ts
```diff
@@ -0,0 +1,72 @@
...
+import { parseVerifiedAnswerYaml } from "../parseVerifiedAnswersYaml";
+import { importVerifiedAnswers } from "../importVerifiedAnswers";
+
+import "dotenv/config";
+
+const commandModule: CommandModule<
+  Record<string, unknown>,
+  ImportCommandArgs
+> = {
+  command: "pages <path>",
```
nlarew, line 20:
```suggestion
  command: "import <path>",
```

Looks like this may have been left over from ingest?

PR #317, opened 2024-02-01
Update openapi spec w/ streaming behavior

mongodben:
Jira: n/a

## Changes

- Update the OpenAPI spec to describe streaming behavior.
- Minor cosmetic updates.

## Notes

- I wasn't able to get the example for the stream response to render well. Based on googling, seems it's a bug. I think it's acceptable as is.

### DIFF
--- a/docs/docs/server/openapi.yaml
+++ b/docs/docs/server/openapi.yaml
@@ -22,7 +22,7 @@ paths:
   /conversations:
     post:
       operationId: createConversation
-      summary: Start a new conversation
+      summary: Start new conversation
       responses:
         200:
           description: OK
@@ -38,7 +38,8 @@ paths:
   /conversations/{conversationId}/messages:
     post:
       operationId: addMessage
-      summary: Send a message to the conversation
+      summary: Add message to the conversation
+      description: Add message to conversation and get response from chatbot.
       parameters:
         - $ref: "#/components/parameters/conversationId"
         - name: stream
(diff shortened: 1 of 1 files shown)


docs/docs/server/openapi.yaml
```diff
@@ -38,7 +38,8 @@ paths:
   /conversations/{conversationId}/messages:
     post:
       operationId: addMessage
-      summary: Send a message to the conversation
+      summary: Add message to the conversation
+      description: Add message to conversation and get response from chatbot.
```
nlarew, line 42:
```suggestion
      description: Add a message to the conversation and get a response back from chatbot.
```

nit: I think we should use proper sentences for the description since space to display the text isn't really a constraining factor here

docs/docs/server/openapi.yaml
```diff
@@ -62,11 +63,38 @@ paths:
                     Markdown-formatted message from the user.
       responses:
         200:
-          description: OK
+          x-summary: OK
+          description: |
+            If you set the query parameter `?stream=false`, the server responds with `application/json`.
```
nlarew, line 68: Can we indicate in the `parameters` section that `stream` defaults to `false`?

docs/docs/server/openapi.yaml
```diff
@@ -96,6 +124,32 @@ paths:
         500:
           $ref: "#/components/responses/InternalServerError"
 
+  /conversations/{conversationId}/messages/{messageId}/comment:
```
nlarew, line 127: Weird that this endpoint is showing up in the diff even though it's already on `main` :/

docs/docs/server/openapi.yaml
```diff
@@ -164,39 +218,109 @@ components:
...
+            type: string
+            format: uri
+            description: |
+              URL to the reference. For example, "https://mongodb.com/docs/atlas/getting-started".
+    MessageResponseStreamEvent:
+      oneOf:
+        - $ref: "#/components/schemas/StreamEventDelta"
+        - $ref: "#/components/schemas/StreamEventReferences"
+        - $ref: "#/components/schemas/StreamEventFinished"
+      description: Item in data stream when streaming a response message.
```
nlarew, line 264:
```suggestion
      description: A server-sent event data payload for a streaming message.
```

docs/docs/server/openapi.yaml
```diff
@@ -164,39 +218,109 @@ components:
...
+    StreamEvent:
```
nlarew, line 265: I don't think we need this abstract event - we're overwriting all the fields in every instance we use it

nlarew (approved):
A few small tweaks but overall LGTM

PR #322, opened 2024-02-05
(EAI-33): Move cleanup to testHelpers

cbush:
Jira: https://jira.mongodb.org/browse/EAI-33

## Changes

- Moves test database creation and cleanup to beforeAll/afterAll hooks within the actual test helpers function
- Removes need for callers to drop database or close MongoClient

## Notes

- This still won't clean up in the event of an aborted test run, so we should periodically run the cleanup script

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/test/testHelpers.ts
+++ b/packages/chatbot-server-mongodb-public/src/test/testHelpers.ts
@@ -1,16 +1,33 @@
+import { strict as assert } from "assert";
 import {
   MongoClient,
+  Db,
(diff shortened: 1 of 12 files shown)


packages/chatbot-server-mongodb-public/src/test/testHelpers.ts
mongodben, line 1: nice!

mongodben (approved):
nice work. if you include the clean up cron in this PR, i can re-review.

mongodben (approved):

PR #325, opened 2024-02-07
(EAI-223) [UI] Throttle suggested prompts

nlarew:
Jira: (EAI-223) [UI] Throttle suggested prompts

## Changes

- Only runs the `onPromptSelected` callback when `showPrompts` is `true` + set `showPrompts` to `false` on the first click.

### DIFF
--- a/packages/mongodb-chatbot-ui/src/Message.tsx
+++ b/packages/mongodb-chatbot-ui/src/Message.tsx
@@ -317,33 +317,37 @@ export const MessagePrompts = ({
   messagePrompts,
   messagePromptsOnClick,
 }: MessagePromptsProps) => {
-  const [inProp, setInProp] = useState(true);
+  const [showPrompts, setShowPrompts] = useState(true);
   const [suggestedPromptIdx, setSuggestedPromptIdx] = useState(-1);
   const nodeRef = useRef(null);
   const duration = 300;
 
+  const onPromptSelected = (prompt: string, idx: number) => {
+    setSuggestedPromptIdx(idx);
+    setShowPrompts(false);
+    setTimeout(() => {
+      messagePromptsOnClick(prompt);
+    }, duration);
+  };
+
   return (
     <CSSTransition
-      in={inProp}
+      in={showPrompts}
       timeout={duration}
       nodeRef={nodeRef}
       classNames={styles.message_prompts}
     >
       <div className={styles.message_prompts} ref={nodeRef}>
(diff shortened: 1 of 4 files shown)


packages/mongodb-chatbot-ui/src/Message.tsx
```diff
@@ -317,33 +317,37 @@ export const MessagePrompts = ({
...
-              onClick={() => {
-                setSuggestedPromptIdx(idx);
-                setInProp(false);
-                setTimeout(() => {
-                  messagePromptsOnClick(messagePrompts[idx]);
-                }, duration);
-              }}
+              key={suggestedPrompt}
+              onClick={() =>
+                showPrompts === true && onPromptSelected(suggestedPrompt, idx)
```
cbush, line 346: Maybe could use a comment here so we remember later why we did this. Also may as well put the throttling functionality in `onPromptSelected()` so we don't have logic spread across two places.

cbush: I'm mildly  concerned about `showPrompts` as a throttling check variable - since it's a state variable it will take a whole render before it is actually set to false. So any clicks between first click and re-render (however small a window) will still register with `showPrompts` still true?

packages/mongodb-chatbot-ui/src/Message.tsx
```diff
@@ -317,33 +317,37 @@ export const MessagePrompts = ({
...
   const [suggestedPromptIdx, setSuggestedPromptIdx] = useState(-1);
   const nodeRef = useRef(null);
   const duration = 300;
 
+  const onPromptSelected = (prompt: string, idx: number) => {
+    setSuggestedPromptIdx(idx);
+    setShowPrompts(false);
+    setTimeout(() => {
+      messagePromptsOnClick(prompt);
+    }, duration);
```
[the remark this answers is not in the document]
nlarew: Sure easy change

packages/mongodb-chatbot-ui/src/Message.tsx
nlarew: 1. Yeah we can move the check to the implementation of `onPromptSelected()`
2. I _think_  React handles this pretty gracefully but I could misunderstand the rendering/batching process. I think I can change to use a ref value that works independently of the render cycle. Lemme look into it.

cbush (approved):

PR #330, opened 2024-02-09
(EAI-226) [UI] Change Conversations without Reload + [SERVER] get conversation route fix

nlarew:
Jira: (EAI-226) [UI] Change Conversations without Reload

## Changes

- UI
  - `ConversationService`
    - Adds support for the `getConversation` route
  - `useConversation()` hook
    - Adds a `switchConversation()` method to allow switching to a different, existing conversation by ID.
    - Removes a check in `createConversation()` to allow starting a new conversation while one already exists.
- Server
  - fix the get conversations routes so that it doesn't return the system message or non user/assistant response messages. 

## Notes

- This is tough to test because we currently disable the `getConversations` route in our MongoDB server implementation via the [blockGetRequests middleware](https://github.com/mongodb/chatbot/blob/main/packages/chatbot-server-mongodb-public/src/middleware/blockGetRequests.ts).
- To test, you can [comment out this middleware in the config](https://github.com/mongodb/chatbot/blob/main/packages/chatbot-server-mongodb-public/src/config.ts#L140) locally and then add a call to `conversation.switchConversation("65c68cc18726aed95fcf6d73")` somewhere in the code (I suggest adding it to the `shouldClose` handler in `ModalView`).
- Non-MongoDB implementations allow the `getConversations` route by default so this should be handy for them.

### DIFF
--- a/packages/mongodb-chatbot-server/src/routes/conversations/utils.test.ts
+++ b/packages/mongodb-chatbot-server/src/routes/conversations/utils.test.ts
@@ -1,4 +1,11 @@
-import { areEquivalentIpAddresses, isValidIp } from "./utils";
+import { Conversation } from "../../services";
+import {
+  areEquivalentIpAddresses,
+  convertConversationFromDbToApi,
+  convertMessageFromDbToApi,
+  isValidIp,
+} from "./utils";
+import { ObjectId } from "mongodb-rag-core";
 
 describe("Conversation routes utils", () => {
   describe("isValidIp()", () => {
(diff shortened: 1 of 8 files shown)


packages/mongodb-chatbot-server/src/routes/conversations/utils.ts
```diff
@@ -27,26 +27,54 @@ export const ApiConversation = z.object({
 
 export function convertMessageFromDbToApi(message: Message): ApiMessage {
```
mongodben, line 28: can you add test(s) for changes to this func?

packages/mongodb-chatbot-server/src/routes/conversations/utils.ts
```diff
@@ -27,26 +27,54 @@ export const ApiConversation = z.object({
...
+    };
+  }
+  return apiMessage;
+}
+
+function assertNever(x: never): never {
+  throw new Error("Unexpected object: " + x);
 }
+
+export function isMessageAllowedInApiResponse(message: Message) {
```
mongodben, line 51:
```suggestion
function isMessageAllowedInApiResponse(message: Message) {
```
i don't think this is used externally

packages/mongodb-chatbot-server/src/routes/conversations/utils.ts
```diff
@@ -27,26 +27,54 @@ export const ApiConversation = z.object({
...
+    case "function":
+      return false;
+    default:
+      // This should never happen - it means we missed a case in the switch.
+      // The assertNever function raises a type error if this happens.
+      return assertNever(message);
+  }
+}
+
 export function convertConversationFromDbToApi(
```
mongodben, line 69: can you add test(s) for this?

mongodben (approved):
a couple of small additions for the server changes but otherwise LGTM.

also updated the PR description + title to capture server changes

packages/mongodb-chatbot-server/src/routes/conversations/utils.test.ts
```diff
@@ -32,3 +39,137 @@ describe("Conversation routes utils", () => {
...
+      createdAt: new Date("2024-01-01T00:00:45Z"),
+    },
+    {
+      id: new ObjectId("65ca768341f9ea61d048aaa8"),
+      role: "function",
+      name: "getBookRecommendations",
+      content: JSON.stringify([
+        { title: "The Way of Kings", author: "Brandon Sanderson" },
+        { title: "The Name of the Wind", author: "Patrick Rothfuss" },
+        { title: "Uprooted", author: "Naomi Novik" },
```
mongodben, line 78: i feel like some scifi would be more appropriate here...

mongodben (approved):

PR #335, opened 2024-02-13
Update drone paths to include root package(-lock).json changes

cbush:
## Notes

- Added it everywhere since package.json changes are likely to merit rebuild for staging?
-


mongodben (approved):
lgtm, seems reasonable. i don't anticipate any issue

PR #336, opened 2024-02-13
(EAI-232):  Scaffold eval package

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-232

## Changes

Scaffold eval package with the following:

- [x] Project config, like TypeScript, eslint, jest, etc.
- [x] types/interfaces
- [x] Mock implementations of data stores
- [x] Mock implementations of commands

## Notes

- Heavily based on set up from the ingest CLI
- I suspect some aspects of this will change as the project develops, but think useful to have this scaffold to build based off of.

### DIFF
--- a/packages/mongodb-chatbot-eval/src/withConfig.ts
+++ b/packages/mongodb-chatbot-eval/src/withConfig.ts
@@ -91,7 +91,7 @@ export const withConfigOptions = <T>(
     description: "Path to config JS file.",
   });
 };
-
+// NOTE: consider moving the generic configurable CLI stuff to `mongodb-rag-core` or another independent library since we're using across multiple projects.
 /**
   Config with promises resolved.
  */
--- a/packages/mongodb-chatbot-eval/src/withConfig.ts
+++ b/packages/mongodb-chatbot-eval/src/withConfig.ts
@@ -1,0 +1,162 @@
+import yargs from "yargs";
+import Path from "path";
+import { logger } from "mongodb-rag-core";
+import { EvalConfig } from "./EvalConfig";
+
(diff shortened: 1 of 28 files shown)


packages/mongodb-chatbot-eval/src/withConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+  return args.option("config", {
+    string: true,
+    description: "Path to config JS file.",
+  });
+};
+
+/**
+  Config with promises resolved.
+ */
+export type ResolvedConfig = {
```
cbush, line 98: We could make a note to refactor this and move it into core for general use. (Not for this PR)

cbush (approved):

packages/mongodb-chatbot-eval/src/withConfig.ts
mongodben: i think the "configurable CLI" thing could be it's own library, not nec part of core. but yes, will add note

packages/mongodb-chatbot-eval/src/withConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+export const withConfigOptions = <T>(
+  args: yargs.Argv<T>
+): yargs.Argv<T & LoadConfigArgs> => {
+  return args.option("config", {
+    string: true,
+    description: "Path to config JS file.",
+  });
+};
+
```
mongodben, line 94:
```suggestion
// NOTE: consider moving the generic configurable CLI stuff to `mongodb-rag-core` or another independent library since we're using across multiple projects.
```

PR #355, opened 2024-03-04
(EAI-276): Faithfulness evaluation

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-276

## Changes

- Add new evaluator for response faithfulness
- Update the `Report` type to include type and name
- Update the reporter to be generic for all binary-style tests

## Notes

-


packages/mongodb-chatbot-eval/src/evaluate/evaluateConversationFaithfulness.ts
```diff
@@ -0,0 +1,96 @@
...
+      "Invalid data type. Expected 'conversation' data."
+    );
+    const conversationData = generatedData as ConversationGeneratedData;
+    const {
+      data: { messages },
+      evalData: { name },
+    } = conversationData;
+    const ragUserMessage = messages.find(
+      (m) => m.role === "user" && m.contextContent
+    ) as UserMessage;
```
cbush, line 51: Why this `as`?
Missing `| undefined`?

cbush (approved):

packages/mongodb-chatbot-eval/src/evaluate/evaluateConversationFaithfulness.ts
mongodben: yeah we should have the `| undefined`. good catch

`as` to cast `Message` as `UserMessage`, which given the find query it will be.

PR #358, opened 2024-03-05
(EAI-252, EAI-253) Ingest & Chunk Code Files

nlarew:
Jira:
- (EAI-252) [Ingest] Code as text files
- (EAI-253) [Ingest] Language-aware code chunking

## Changes

- Adds `CodeOnGithubTextDataSource` to ingest code from a repo in `txt` format
- Adds a partial implementation of MongoDB Developer repos in `mongodbDeveloperCodeRepos`
- Adds language-aware chunking for everything supported by langchain

## Notes

- The repos in `mongodbDeveloperCodeRepos` are mainly just an example for now. We don't actually include them in our list of sources to ingest.
- There is a lot of config if we want to include all these repos. We should instead consider a pull-based approach e.g. https://jira.mongodb.org/browse/EAI-270


packages/ingest-mongodb-public/src/sources/mongodbDeveloperCodeRepos.test.ts
mongodben, line 1: can you do similar smoke tests for each repo to make sure that we don't loose them?

packages/mongodb-rag-ingest/src/sources/CodeOnGithubTextDataSource.ts
```diff
@@ -0,0 +1,88 @@
...
+        /\.gcloudignore/,
+        /\.editorconfig/,
+        /\.vscode/,
+        ...(repoLoaderOptions?.ignoreFiles ?? []),
+      ],
+    },
+    async handleDocumentInRepo(document) {
+      const page: Page = {
+        body: document.pageContent,
+        format: "txt",
```
mongodben, line 47: why not let this be passable to account for different formats of different languages? 

`java`, `js`, etc

mongodben: supporting "intelligent" chunking on a variety of different languages shouldn't be too hard using the langchain `RecursiveCharacterTextSplitter.fromLanguage()` method.

mongodben:
> The repos in mongodbDeveloperCodeRepos are mainly just an example for now. We don't actually include them in our list of sources to ingest.

what are you thinking for next steps here @nlarew?

mongodben (changes requested):
nice work. see comments. most important thing in my mind is why use the `txt` format instead of for each page, use whatever the file extension is of that page?

packages/ingest-mongodb-public/src/sources/mongodbDeveloperCodeRepos.test.ts
nlarew: Yup added one for all 6 configured repos

packages/mongodb-rag-ingest/src/sources/CodeOnGithubTextDataSource.ts
nlarew: I went a little crazy on this one and added some machinery around a list of "canonical" page formats (and their synonyms). It's a superset of the existing type and should be relatively flexible, though maybe I should add a way to append to it from the ingest config 🤔

nlarew:
> > The repos in mongodbDeveloperCodeRepos are mainly just an example for now. We don't actually include them in our list of sources to ingest.
> 
> what are you thinking for next steps here @nlarew?

@mongodben Once the broader implementation here is LGTM'd I'd like to run an A/B test with our quality evaluation to see if/how the raw code chunks help

packages/mongodb-rag-core/src/PageFormat.ts
```diff
@@ -0,0 +1,99 @@
...
+  ["kotlin", "kt"],
+  ["latex", "tex"],
+  ["objective-c", "m"],
+  ["php", "php"],
+  ["python", "py"],
+  ["ruby", "rb"],
+  ["rust", "rs"],
+  ["scala", "sc"],
+  ["shell", "sh"],
+  ["swift", "swift"],
```
mongodben, line 44: also can be just 1 elem, no?

packages/mongodb-rag-ingest/src/embed/chunkCode.ts
```diff
@@ -0,0 +1,118 @@
...
+        : preTransformChunk;
+
+      const chunk = {
+        ...transformedChunk,
+        tokenCount: tokenizer.encode(transformedChunk.text).bpe.length,
+      };
+      // const { metadata } = extractFrontMatter(transformedChunk.text);
+      // if (metadata) {
+      //   chunk["metadata"] = metadata;
+      // }
```
mongodben, line 58:
```suggestion
```

packages/mongodb-rag-ingest/src/embed/chunkCode.ts
```diff
@@ -0,0 +1,118 @@
...
+  scala: "scala",
+  swift: "swift",
+  latex: "latex",
+  html: "html",
+};
+
+export function isSupportedLanguage(
+  str: PageFormat
+): str is LangchainSupportedCodePageFormat {
+  return str in supportedLanguageMap;
```
mongodben, line 108: what's the `in` operator here?

mongodben (approved):
LGTM.

next typescript level unlocked on this one

packages/mongodb-rag-ingest/src/embed/chunkCode.ts
nlarew: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

In this case it's equivalent to

```js
Object.keys(supportedLanguageMap).includes(str)
```

PR #375, opened 2024-03-20
(EAI-24): Expected links evaluation

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-24

## Changes

- Add evaluation for if generated conversation contains certain links.
- Include `expectedLinks` in conversation generated data
- Add implementation of if top search queries include the expected links.
  - The data for this, stored in comes`packages/chatbot-eval-mongodb-public/testCases/included_links_conversations.yaml`, comes from https://github.com/mongodb/docs-search-transport/blob/main/src/data/term-result-mappings.ts

## Notes

-


packages/chatbot-eval-mongodb-public/testCases/included_links_conversations.yml
```diff
@@ -0,0 +1,832 @@
+- name: Search for 'and'
+  tags:
+    - top_search_term
+  messages:
+    - role: user
+      content: and
+  expectation: "N/A"
```
nlarew, line 7: All these `expectation: "N/A"` make me think we should tweak the data model to make `expectation` optional at some level. Thoughts?

packages/mongodb-chatbot-evaluation/src/evaluate/evaluateExpectedLinks.test.ts
```diff
@@ -0,0 +1,201 @@
...
+        expectedLinksFound: {
+          bar: {
+            matchingActualLink: "bar",
+            includesExpected: true,
+          },
+          fizz: {
+            matchingActualLink: "fizz_buzz",
+            includesExpected: true,
+          },
+        },
```
nlarew, line 72: I think we could tweak the schema a bit here.

1. It's a bit weird to have links we expected but didn't find under `expectedLinksFound`
2. I think `includesExpected` is effectively the same as `matchingActualLink !== undefined`. If so we can make this implicit in the data model.

Maybe something like this?

```ts
interface Metadata {
  expectedLinks: {
    // Map of expectedLinks to one or more matching links we found
    matched: { [expectedLink: string]: [string, ...string[]] };
    // List of links we expected but didn't find
    unmatched: string[];
  };
}
```

So this test would be:

```ts
{
  metadata: {
    testName: generatedConversationData.evalData.name,
    expectedLinks: {
      matched: {
        bar: ["bar"],
        fizz: ["fizz_buzz"],
      },
      unmatched: [],
    },
  },
}
```

packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - reference/operator/aggregation/or
+    - reference/operator/query-logical
+- name: Search for 'not'
+  tags:
+    - top_search_term
+  messages:
+    - role: user
+      content: not
+  expectation: "N/A"
+  # no expected links
```
nlarew, line 30: Could we allow `expectedLinks: null` or `expectedLinks: []` to explicitly define this use case? I think we'd have to refine our implementation slightly.

packages/mongodb-chatbot-evaluation/src/evaluate/evaluateExpectedLinks.test.ts
mongodben: seems reasonable. will adjust

packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml
mongodben: yeah we can make the construct a little more flexible.

packages/chatbot-eval-mongodb-public/testCases/included_links_conversations.yml
mongodben: yea we can do that. will need to touch some other files as well to take into account the optionality. PR will grow a bit.

packages/mongodb-chatbot-evaluation/src/evaluate/evaluateExpectedLinks.test.ts
mongodben: w/ a bit more thought on it, went with:

```ts
export interface ExpectedLinks {
  /**
    The expected link is the key and the value is an array
    of actual links that match the expected link.
   */
  [expectedLink: string]: string[];
}
```

and array is empty if no expected links. bit of a cleaner API

nlarew (approved):
LGTM overall thanks for the changes. Couple Qs on the test data but nothing blocking

packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml
mongodben: i did!

packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml
```diff
@@ -0,0 +1,30 @@
+- name: Search for 'and'
+  tags:
+    - top_search_term
+  messages:
+    - role: user
+      content: and
+  expectation: "N/A"
```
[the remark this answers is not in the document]
mongodben: oh i guess i forgot to remove the N/As! will do before merge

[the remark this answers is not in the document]
mongodben: ok so this is odd. still seeing the N/As here in the PR but i definitely removed them previously in the commit [abec355](https://github.com/mongodb/chatbot/pull/375/commits/abec3556c401fda71972a0ddbb18fa16a2ce2663)

see current HEAD of branch EAI-24 https://github.com/mongodb/chatbot/blob/EAI-24/packages/chatbot-eval-mongodb-public/testCases/included_links_conversations.yml#L1

you ever seen this before?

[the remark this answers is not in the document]
nlarew: You linked to a different file - the N/A's here are still present on the HEAD of the branch: https://github.com/mongodb/chatbot/blob/EAI-24/packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml

packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml
nlarew: I guess in this case I'd expect to see `expectedLinks: []` so that we're explicitly saying "there should be no related links" as opposed to "we're not testing for related links"

packages/mongodb-chatbot-evaluation/testData/expected_links_tests.yaml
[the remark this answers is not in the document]
mongodben: ah derp. that'd explain it.

nlarew (approved):

PR #377, opened 2024-03-21
Update tck

cbush:


nlarew (approved):
LGTM

Might be nice if we could define this in one place (e.g. as an arg to the CLI command) because it's gonna get annoying to have all of these once our corpus grows

cbush:
Indeed, captured here: https://jira.mongodb.org/browse/EAI-296

PR #383, opened 2024-03-27
(EAI-243): Lint space between entities

cbush:
Jira: https://jira.mongodb.org/browse/EAI-243

## Changes

- Adds eslint rules to enforce line breaks between entities (classes, functions, class members) 
- Adds lint:fix script to most packages

## Notes

- You might want to set up eslint fix as your auto-formatter on save (search "Format Code Action"): https://blog.logrocket.com/using-prettier-eslint-automate-formatting-fixing-javascript/
- Reviewer: most of the file changes are highly superficial - spot check only


mongodben (approved):

.prettierrc.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  embeddedLanguageFormatting: "off",
```
nlarew, line 2: Was there a reason to disable this? Seems like it might be marginally useful in README, CONTRIBUTING, etc. docs

nlarew (approved):

.prettierrc.cjs
cbush: Oh interesting - I guess prettier does .md? The reason I turned it off is because we have some html string literals that are getting compressed. See [packages/mongodb-artifact-generator/src/operations.ts](https://github.com/mongodb/chatbot/blob/main/packages/mongodb-artifact-generator/src/operations.ts#L87-L109) for example: this would strip the newlines and put everything into one block.

nlarew: Ah very annoying there so probably a good change

cbush: Also I'm not sure that prettier is what handles the formatting within md files. I wasn't able to get embedded languages formatted either way.

PR #384, opened 2024-03-27
(EAI-245): Create IDs for categories of questions to track FAQ across…

cbush:
… days

Jira: https://jira.mongodb.org/browse/EAI-245

## Changes

- Updates the findFaq script to find a 'faqId' from a previous similar entry. If one can't be found, creates a new one.
- Optionally backports new IDs to previous entries.

## Notes

- This will enable us to bucket FAQ reports by faqId, which will provide more insight into frequently asked questions.

### DIFF
--- a/packages/scripts/src/findFaq.ts
+++ b/packages/scripts/src/findFaq.ts
@@ -1,5 +1,5 @@
 import { strict as assert } from "assert";
-import { Db, ObjectId } from "mongodb";
+import { Db, ObjectId, Collection, WithId } from "mongodb";
 
 import {
   Conversation,
@@ -7,6 +7,9 @@ import {
   AssistantMessage,
   FunctionMessage,
   UserMessage,
+  VectorStore,
+  FindNearestNeighborsOptions,
+  WithScore,
 } from "mongodb-chatbot-server";
 import { clusterize, DbscanOptions } from "./clusterize";
 import { findCentroid } from "./findCentroid";
@@ -50,6 +53,11 @@ export type FaqEntry = {
     (a cluster with more objects in it is a more frequently asked question).
    */
   faqScore: number;
+
+  /**
+    An id unique to this category of questions.
+   */
+  faqId?: string;
 };
 
 export const findFaq = async ({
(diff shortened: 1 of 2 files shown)


packages/scripts/src/findFaq.ts
```diff
@@ -166,3 +174,127 @@ export const findFaq = async ({
...
+              {
+                _id: q._id,
+              },
+              { $set: { faqId } }
+            );
+            console.log(
+              `Backported new ID ${faqId} to ${q._id} (modified ${updateResult.modifiedCount})`
+            );
+          })
+        );
```
nlarew, line 295: This might be a good opportunity for `bulkWrite` since this could fire off up to `k` db calls per FAQ cluster. Probably not a big deal at our scale but something to consider
```suggestion
        await faqCollection.bulkWrite(
          previousFaqsWithoutIds.map((q) => ({
            updateOne: {
              filter: { _id: q._id },
              update: { $set: { faqId } },
            },
          }))
        );
```

packages/scripts/src/main/findFaqMain.ts
```diff
@@ -30,26 +35,43 @@ async function main() {
...
+      clusterizeOptions: {
         epsilon,
-        created,
-        _id: new ObjectId(),
-      }))
-    );
-    console.log(`Inserted ${insertResult.insertedCount} FAQ entries.`);
+      },
+    })
+      .then(async (faq) => {
```
nlarew, line 52: Any reason to use a promise chain here instead of `await`?

packages/scripts/src/findFaq.ts
```diff
@@ -166,3 +174,127 @@ export const findFaq = async ({
...
+  };
+};
+
+/**
+  For each given question, finds if any similar messages already have a faqId.
+  If so, adopts the faqId. Otherwise, invents a new faqId for this category of
+  question.
+ */
+export const assignFaqIds = async ({
+  faq,
```
nlarew, line 242:
```suggestion
  faqs,
```

```suggestion
  faqEntries,
```

Thoughts? I think it's useful to have `faq` => one cluster of questions and `faqs` => `faq[]`.

Would entail some small updates throughout

packages/scripts/src/findFaq.ts
```diff
@@ -166,3 +174,127 @@ export const findFaq = async ({
...
+        } faqId ${faqId} for question category "${q.question}"`
+      );
+
+      if (backportNewIds) {
+        const previousFaqsWithoutIds = previousFaqs.filter(
+          (q) => q.faqId === undefined
+        );
+        await Promise.all(
+          previousFaqsWithoutIds.map(async (q) => {
+            const updateResult = await faqCollection.updateOne(
```
nlarew, line 285: Can we refactor to move this db mutation out of `assignFaqIds()` and into the same step as inserting the new FAQs in `findFaqMain`?

I think it's best to keep this as a pure function that returns a "draft" of the new & updated FAQs and then later we can upsert them all together.

packages/scripts/src/main/findFaqMain.ts
cbush: Yes (you know I would never use a promise chain without a reason).

In this case I wanted each scope to be self-contained rather than assigning some variable over and over or having a bunch of `await Promise.all()`s. So I'm using the Promise.then() as a pipeline.

packages/scripts/src/findFaq.ts
```diff
@@ -166,3 +174,127 @@ export const findFaq = async ({
...
+    assigns the new faqId to the existing entries. 
+   */
+  backportNewIds?: boolean;
+}): Promise<(FaqEntry & { faqId: string })[]> => {
+  const store = makeFaqVectorStoreCollectionWrapper(faqCollection);
+
+  return await Promise.all(
+    faq.map(async (q) => {
+      // See if there already is an ID for this FAQ.
+      const previousFaqs = await store.findNearestNeighbors(q.embedding);
```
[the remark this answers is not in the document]
cbush: It is not currently possible, no

packages/scripts/src/findFaq.ts
```diff
@@ -166,3 +174,127 @@ export const findFaq = async ({
...
+}: {
+  faq: FaqEntry[];
+  faqCollection: Collection<
+    WithId<FaqEntry & { created: Date; epsilon: number }>
+  >;
+
+  /**
+    If a new faqId is generated because existing entries did not have a faqId,
+    assigns the new faqId to the existing entries. 
+   */
```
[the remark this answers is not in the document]
cbush: Good point!

packages/scripts/src/findFaq.ts
cbush: It's a really great point and I did that refactor, but it ended up making the function do too much, and I realized that that the "backport" feature was overbaked, so I took it out completely.

nlarew: SGTM. I think a "faq without a faqId" is only relevant for our existing FAQs from before this merges anyway so probably not a big deal long term

nlarew (approved):
LGTM - failing test is a timeout error in an unrelated package

PR #389, opened 2024-03-28
(EAI-235): Publish evaluation CLI package

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-235

## Changes

- Add drone pipeline to publish eval CLI
- I also published eval package `0.1.0`

## Notes

-


.drone.yml
```diff
@@ -871,3 +871,31 @@ steps:
...
+    - tag
+  ref:
+    include:
+      - refs/tags/mongodb-chatbot-evaluation-v*
+
+steps:
+  - name: build package
+    image: node:18
+    commands:
+      - npm install lerna
```
nlarew, line 891:
```suggestion
      - npm install
```
We may add other dependencies (e.g. we just added `@stylistic/eslint-plugin-js`) should we default to installing everything?

nlarew (approved):
LGTM w/ one question

.drone.yml
mongodben: that's covered in the following `npm run bootstrap` step. it runs install

mongodben: though if we upgrade lerna to latest, i believe they deprecated `bootstrap`, so we could refactor then.

Issue #382, opened 2024-03-27
Use Claude3 LLM


eric-gardyn:
I wanted to test with Claude3 LLM.
Out-of-the-box, framework didn't work; just wanted to share my changes:

I had to change LangchainChatLlm.ts for answerQuestionAwaited, with

```
...
      const prompts = ChatPromptTemplate.fromMessages(
        messages.map(m => messageBaseMessagePromptTemplateLike(m))
      )
      const chain = prompts.pipe(chatModel)
      const res = await chain.invoke({})
...

function messageBaseMessagePromptTemplateLike(
  message: OpenAiChatMessage
): BaseMessagePromptTemplateLike {
  return [message.role, message.content ?? '']
}

```

mongodben:
thanks eric. i also want to test with claude 3, so will keep this in mind 😄 

 are you using claude from anthropic directly or another provider like bedrock?

eric-gardyn:
directly from Anthropic; this is just for a comparison test

mongodben:
got it, will help me repro the issue. 

maybe while at it, will try it out in our eval suite to see comparison

eric-gardyn:
btw, is the eval suite ready to be used? I am curious to try it

mongodben:
yes it's available on npm and works, but the docs need expansion https://mongodb.github.io/chatbot/evaluation/. i'm planning on working on this over the next week or so. 

our eval suite can be found here - https://github.com/mongodb/chatbot/tree/main/packages/chatbot-eval-mongodb-public

copying the patterns from here is probably the quickest way to get started.

let me know if you'd like to do a call to get set up onboarded w/ it. you can reach out on linkedin for that.

PR #392, opened 2024-03-29
`LangchainChatLlm` support for Anthropic

mongodben:
Jira: n/a

## Changes

- Modify langchain chatllm implementation to support Anthropic models

## Notes

- i believe this change is compensating for a bug on langchain's side, as we were already conforming to their interface
- skipping test for anthropic b/c MongoDB doesn't have an Anthropic subscription that we could use in the integration test here. 
  - You can sign up for free to anthropic and plug in your own API key (that's what i did)
- Thank you to @eric-gardyn for help with the solution implementation!

Resolves https://github.com/mongodb/chatbot/issues/382

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4177,6 +4177,76 @@
       "version": "1.1.1",
       "license": "MIT"
     },
+    "node_modules/@langchain/anthropic": {
+      "version": "0.1.9",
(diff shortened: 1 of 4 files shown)


nlarew (approved):
LGTM

PR #394, opened 2024-04-04
Support LangChain `Embeddings` class

mongodben:
Jira: n/a

## Changes

- `makeLangChainEmbedder()` constructor of `Embedder` that wraps LangChain `Embeddings` class
- document `makeLangChainEmbedder()`

## Notes

- the langchain `Embeddings` class has some nice support for things like retry and caching. we might want to consider migrating to using this instead of Azure OpenAI directly. 
- I want to add this before a hackathon on saturday so folks can use our framework w/ whatever embedding model they choose.


cbush (approved):
🤠

PR #404, opened 2024-04-23
[DOP-4515]: Add ID property to error div

branberry:
Jira: https://jira.mongodb.org/browse/DOP-4515

## Changes

- Move `Modal` behind ternary for the `ModalView` component
- Only provide a value for `initialFocus` if there is no error with the chatbot
- Add config for bundling a CommonJS compatible output

## Notes

- This prevents an error that stems when the Chatbot cannot connect to the server. The `initialFocus` of the LG modal component will break due to the error message displayed for the Chatbot not having the ID associated with `initialFocus`
- I included bundling for CommonJS to allow for our Jest tests to pass. Unfortunately, Jest does not support ESM (not easily, at least), and this leads to test failures in [Snooty](https://github.com/mongodb/snooty/actions/runs/8804003128/job/241634055100).

### DIFF
--- a/packages/mongodb-chatbot-ui/src/ModalView.tsx
+++ b/packages/mongodb-chatbot-ui/src/ModalView.tsx
@@ -48,7 +48,7 @@ export function ModalView(props: ModalViewProps) {
   const { darkMode } = useDarkMode(props.darkMode);
   const { className, inputBarId, ...chatWindowProps } = props;
 
-  const { closeChat, open } = useChatbotContext();
+  const { closeChat, open, conversation } = useChatbotContext();
 
   const shouldClose = () => {
     if (props.shouldClose?.() ?? true) {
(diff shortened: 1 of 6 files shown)


packages/mongodb-chatbot-ui/src/ModalView.tsx
```diff
@@ -62,18 +62,20 @@ export function ModalView(props: ModalViewProps) {
...
-      <Suspense fallback={null}>
-        {open ? (
+    <Suspense fallback={null}>
+      {open ? (
+        <Modal
+          className={cx(styles.modal_container({ darkMode }), className)}
+          open={open}
+          size="large"
+          initialFocus={
+            !conversation.error ? `#${chatWndowInputBarId}` : undefined
```
cbush, line 72: While you're in here, could you fix this in the various places?
```suggestion
            !conversation.error ? `#${chatWindowInputBarId}` : undefined
```

cbush (approved):

packages/mongodb-chatbot-ui/vite.config.component.cjs.ts
```diff
@@ -0,0 +1,52 @@
...
+        Buffer: true,
+        global: true,
+        process: true,
+      },
+      protocolImports: true,
+    }),
+    react(),
+  ],
+  build: {
+    outDir: "build/cjs",
```
branberry, line 25: This is essentially what the `vite.config.component.ts` file was before code splitting was introduced, save for a few things like the `outDir` and `formats: ["cjs"],`

cbush (approved):

PR #409, opened 2024-04-30
Work around test that times out

nlarew:
```
FAIL  src/sources/MongoDbDotComDataSource.test.ts (26.539 s)
  ● MongoDbDotComDataSource › should fetch pages

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      13 | });
      14 | describe("MongoDbDotComDataSource", () => {
    > 15 |   it("should fetch pages", async () => {
         |   ^
      16 |     const pages = await dotcomDataSource.fetchPages();
      17 |     // writeFileSync('pages.json', JSON.stringify(pages, null, 2))
      18 |     expect(pages.length).toBeGreaterThan(0);

      at src/sources/MongoDbDotComDataSource.test.ts:15:3
      at Object.<anonymous> (src/sources/MongoDbDotComDataSource.test.ts:14:1)
```

## Changes

- Adds a `limit` arg to the MongoDbDotCom data source
- Sets the `limit` to 10 so the test doesn't time out

### DIFF
--- a/packages/ingest-mongodb-public/src/sources/MongoDbDotComDataSource.ts
+++ b/packages/ingest-mongodb-public/src/sources/MongoDbDotComDataSource.ts
@@ -11,16 +11,18 @@ import striptags from "striptags";
 export function makeMongoDbDotComDataSource({
   connectionUri,
   dbName,
+  limit,
 }: {
   connectionUri: string;
   dbName: string;
+  limit?: number; // for testing
 }): DataSource {
   return {
     name: "mongodb-dot-com",
     async fetchPages() {
       const mongodb = new MongoClient(connectionUri);
       try {
-        const customerPages = await mongodb
+        const query = mongodb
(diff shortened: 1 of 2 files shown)


packages/ingest-mongodb-public/src/sources/MongoDbDotComDataSource.ts
```diff
@@ -11,16 +11,18 @@ import striptags from "striptags";
 export function makeMongoDbDotComDataSource({
   connectionUri,
   dbName,
+  limit,
 }: {
   connectionUri: string;
   dbName: string;
+  limit?: number; // for testing
```
cbush, line 18: Limit what?

cbush (approved):

PR #427, opened 2024-06-03
Fix missing search filter error

nlarew:


mongodben (approved):
LGTM, let's just make sure that we have all tests passing in the CI before merge.

PR #435, opened 2024-06-11
(EAI-263) Document verified answers

nlarew:
Jira: (EAI-263) Document verified answers

## Changes

- Adds docs for verified answers & FAQ Finder


cbush (approved):
Thanks for doing this!

PR #439, opened 2024-06-17
(EAI-403) configurable collection name for EmbeddedContentStore & PageStore

nlarew:
Jira: (EAI-403) configurable collection name for EmbeddedContentStore & PageStore

### DIFF
--- a/packages/mongodb-rag-core/src/MongoDbEmbeddedContentStore.test.ts
+++ b/packages/mongodb-rag-core/src/MongoDbEmbeddedContentStore.test.ts
@@ -3,7 +3,10 @@ import { strict as assert } from "assert";
 import { DatabaseConnection } from "./DatabaseConnection";
 import { EmbeddedContentStore } from "./EmbeddedContent";
 import { FindNearestNeighborsOptions } from "./VectorStore";
-import { makeMongoDbEmbeddedContentStore } from "./MongoDbEmbeddedContentStore";
+import {
+  type MongoDbEmbeddedContentStore,
+  makeMongoDbEmbeddedContentStore,
+} from "./MongoDbEmbeddedContentStore";
 import { assertEnvVars } from "./assertEnvVars";
 import { CORE_ENV_VARS } from "./CoreEnvVars";
 import { makeOpenAiEmbedder } from "./OpenAiEmbedder";
@@ -21,7 +24,7 @@ const {
 } = assertEnvVars(CORE_ENV_VARS);
 
 describe("MongoDbEmbeddedContentStore", () => {
-  let store: (DatabaseConnection & EmbeddedContentStore) | undefined;
+  let store: MongoDbEmbeddedContentStore | undefined;
   beforeEach(async () => {
     // Need to use real Atlas connection in order to run vector searches
     const databaseName = `test-database-${Date.now()}`;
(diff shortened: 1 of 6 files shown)


packages/mongodb-rag-core/src/MongoDbPageStore.ts
```diff
@@ -14,24 +14,37 @@ export type MongoDbPageStore = DatabaseConnection &
...
       args?: LoadPagesArgs<Filter<PersistedPage>>
     ): Promise<PersistedPage[]>;
+    metadata: {
+      databaseName: string;
+      collectionName: string;
+    };
   };
 
+export type MakeMongoDbPageStoreParams = MakeMongoDbDatabaseConnectionParams & {
+  collectionName?: string;
```
mongodben, line 24: can you add a tsdoc comment about how default is "pages"

smthn like:

```suggestion
  /**
    @default "pages"
   */
  collectionName?: string;
```

mongodben (approved):
lgtm, see suggestions about the tsdoc comments

mongodben:
also, pls fix CI issue before merge

packages/mongodb-rag-core/src/MongoDbEmbeddedContentStore.test.ts
```diff
@@ -206,4 +209,23 @@ describe("nearest neighbor search", () => {
...
+        connectionUri: MONGODB_CONNECTION_URI,
+        databaseName: store.metadata.databaseName,
+        collectionName: "custom-embedded_content",
+      }
+    );
+
+    expect(storeWithCustomCollectionName.metadata.collectionName).toBe(
+      "custom-embedded_content"
+    );
+    await storeWithCustomCollectionName.drop();
```
mongodben, line 229: oopsies!!

nlarew, line 229:
```suggestion
```

packages/mongodb-rag-core/src/MongoDbEmbeddedContentStore.test.ts
```diff
@@ -118,7 +121,7 @@ describe("nearest neighbor search", () => {
     minScore: 0.9,
   };
 
-  let store: (DatabaseConnection & EmbeddedContentStore) | undefined;
+  let store: MongoDbEmbeddedContentStore | undefined;
   beforeEach(async () => {
     // Need to use real Atlas connection in order to run vector searches
     const databaseName = MONGODB_DATABASE_NAME;
```
mongodben, line 127: make this a randomly generated name please

packages/mongodb-rag-core/src/MongoDbEmbeddedContentStore.test.ts
mongodben: seems like this is dropping the whole DB when it should be just the collection

packages/mongodb-rag-core/src/MongoDbEmbeddedContentStore.test.ts
nlarew: unfortunately we can't here - its testing vector search so it needs to point to the db that has the embedded data. Best course is to just not call `db.drop()` (lol) or maybe add a readonly version of the store and use that in these tests

PR #445, opened 2024-07-03
(EAI-354): Discovery eval

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-354

## Changes

- Add set of synthetically generated questions that don't mention mongodb to see if it's discovered in the answer.
- Add regex-based conversation evaluator
- Add discovery evaluation suite
- Support concurrency on LLM conversation data generation
- Support concurrency on evaluation stage

## Notes

-


packages/model-eval/src/discovery.config.ts
```diff
@@ -0,0 +1,73 @@
...
+import fs from "fs";
+import path from "path";
+import "dotenv/config";
+import { MongoClient, assertEnvVars } from "mongodb-rag-core";
+import { envVars } from "./envVars";
+import { makeChatLlmConversationEvalCommands } from "./makeChatLlmEvalCommands";
+import { makeRadiantChatLlm } from "./makeRadiantChatLlm";
+import { radiantModels } from "./radiantModels";
+
+export default async () => {
```
nlarew, line 18: Seems like there's a good deal of overlap here with `quizEval.config.ts` - maybe worth sharing a common base config?

```ts
{
  ...baseModelEvalConfig,
  commands: makeChatLlmConversationEvalCommands({
    chatLlmConfigs,
    testCases: discoveryConversationTestCases,
  }),
}
```

or maybe even combining into a single config, e.g.

```ts
{
  commands: {
    ...makeChatLlmConversationEvalCommands({
      chatLlmConfigs,
      testCases: discoveryConversationTestCases,
    }),
    ...makeChatLlmQuizEvalCommands({
      configs: modelsConfig,
      quizQuestions: quizQuestionTestCases,
    })
  }
}
```

packages/model-eval/src/makeChatLlmEvalCommands.ts
```diff
@@ -0,0 +1,136 @@
...
+    generate: generateConfig,
+    evaluate: evaluationConfig,
+    report: reportConfig,
+  } satisfies EvalConfig["commands"];
+  return commands;
+}
+
+export interface ChatLlmConversationsEvalConfig {
+  name: string;
+}
```
nlarew, line 74: Do we use this anywhere?

packages/model-eval/src/makeChatLlmEvalCommands.ts
```diff
@@ -0,0 +1,136 @@
...
+interface MakeChatLlmConversationEvalCommandsParams {
+  chatLlmConfigs: {
+    chatLlm: ChatLlm;
+    name: string;
+  }[];
+  testCases: ConversationTestCase[];
+}
+
+export function makeChatLlmConversationEvalCommands({
```
nlarew, line 83: IMO it would be nice to have these as separate files, since they're independent evaluations that we might edit separately. Maybe as we add more command factories like this we can refactor a bit.

nlarew (approved):
LGTM w/ non-blocking suggestions

packages/model-eval/src/makeChatLlmEvalCommands.ts
mongodben: looks like no, removing. prob vestige of some previous version

packages/model-eval/src/discovery.config.ts
mongodben: good point. will make a common base

PR #449, opened 2024-07-22
(EAI-457) [UI] Custom chatbot trigger helpers

nlarew:
Jira: (EAI-457) [UI] Custom chatbot trigger helpers

## Changes

- Refactors most of `InputBarTrigger` data and logic into a separate, re-usable hook
- Cleans up unnecessary `classnames` dependency in favor of Emotion's basically equivalent `cx`

## Notes

- Makes it easier for devs to make input bar triggers like ours but with custom components/logic
- Most applicable use is for upcoming DOP work on a new trigger

### DIFF
--- a/packages/mongodb-chatbot-ui/src/FloatingActionButtonTrigger.tsx
+++ b/packages/mongodb-chatbot-ui/src/FloatingActionButtonTrigger.tsx
@@ -1,5 +1,4 @@
-import { css } from "@emotion/css";
-import classNames from "classnames";
+import { css, cx } from "@emotion/css";
 import {
   ActionButtonTrigger,
   type ActionButtonTriggerProps,
@@ -30,7 +29,7 @@ export function FloatingActionButtonTrigger({
 }: FloatingActionButtonTriggerProps) {
   return (
     <ActionButtonTrigger
-      className={classNames(styles.chat_trigger, className)}
+      className={cx(styles.chat_trigger, className)}
       {...props}
     />
   );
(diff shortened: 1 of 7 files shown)


packages/mongodb-chatbot-ui/src/FloatingActionButtonTrigger.tsx
```diff
@@ -1,5 +1,4 @@
-import { css } from "@emotion/css";
-import classNames from "classnames";
+import { css, cx } from "@emotion/css";
```
mongodben, line 1: why these changes? can you note why in the PR description?

packages/mongodb-chatbot-ui/src/useTextInputTrigger.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export type UseTextInputTriggerArgs = {
+  placeholder?: string;
+  fatalErrorMessage?: string;
+};
+
+/**
+  The base props for a Chatbot trigger component that allows the user to input text.
+ */
+export type ChatbotTextInputTriggerProps = ChatbotTriggerProps &
```
mongodben, line 15: do we need to update the public facing docs for this PR?

packages/mongodb-chatbot-ui/src/FloatingActionButtonTrigger.tsx
nlarew: Basically just tidying up - we don't need this lib because emotion has their own version baked in

packages/mongodb-chatbot-ui/src/useTextInputTrigger.ts
nlarew: Nothing required, though we could cover custom triggers/view more generally in the docs & mention this there.

https://jira.mongodb.org/browse/EAI-468

PR #459, opened 2024-07-30
(EAI-472) FAQ Finder script is missing search filter

nlarew:
Jira: (EAI-472) FAQ Finder script is missing search filter

## Changes

- Adds a default `filter = {}` for vector search to avoid the error breaking the script


mongodben (approved):
assuming this is the same fix that we applied to the other places

PR #460, opened 2024-07-30
(EAI-447) Add Prometheus alerts for runtime server errors

nlarew:
Jira: (EAI-447) Add Prometheus alerts for runtime server errors

## Changes

- We should now get email alerts if the chatbot is responding 5XX to more than 2% of all requests in a 5 minute period

### DIFF
--- a/packages/chatbot-server-mongodb-public/environments/production.yml
+++ b/packages/chatbot-server-mongodb-public/environments/production.yml
@@ -81,3 +81,15 @@ probes:
 defaultAlerts:
   enabled: true
   email: "chatbot-alerts@mongodb.com"
+
+prometheusRules:
+  - alert: TooManyHttpInternalServerErrors
+    # (Total 500 Requests / Total Requests) > X%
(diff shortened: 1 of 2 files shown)


packages/chatbot-server-mongodb-public/environments/production.yml
```diff
@@ -81,3 +81,15 @@ probes:
 defaultAlerts:
   enabled: true
   email: "chatbot-alerts@mongodb.com"
+
+prometheusRules:
```
mongodben, line 85: can you also add this to staging server? 

and also, would be good to test out on staging somehow...maybe temporarily break by changing an env var, then run a non-small load against the server to make sure it works. perhaps could do in the `performance-test` directory, which uses a load testing framework to make this easy

mongodben (changes requested):
pls see comment

packages/chatbot-server-mongodb-public/environments/production.yml
nlarew: > can you also add this to staging server?

Yup - added in this PR just now

> would be good to test out on staging somehow

I think to get the staging environment to actually use it I'll need to push the updated staging config to `main`. Then I can point a test suite at staging and verify. Just noting the plan here + seeing if you have any alternative ideas.

PR #461, opened 2024-08-05
(EAI-379): Add more radiant models

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-379

## Changes

- Add additional Radiant models to config
- Add test that validates all models are working.
- Rename `model-eval` package to `benchmarks` for clarity on package purpose + uniformity w/ other naming

## Notes

-


packages/benchmarks/src/radiantModels.test.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+import { assertEnvVars } from "mongodb-rag-core";
+import { envVars } from "./envVars";
+import { makeRadiantChatLlm } from "./makeRadiantChatLlm";
+import { radiantModels } from "./radiantModels";
+
+jest.setTimeout(60000);
+// NOTE: due to this issue https://github.com/nodejs/node/issues/39964,
+// you must run the tests with a Node version >= 20.0.0
+describe.skip("Radiant models", () => {
```
nlarew, line 13: Test won't do us much good if we're skipping it. Can we track a follow up for this as an EAI?

packages/benchmarks/tsconfig.json
```diff
@@ -1,3 +1,10 @@
+// {
+//   "extends": "../../tsconfig.json",
+//   "compilerOptions": {
+//     "outDir": "./build"
+//   },
+//   "include": ["./src/**/*.ts"]
+// }
```
nlarew, line 7:
```suggestion
```

nlarew (approved):
Changes LGTM, 1 follow up Q

packages/benchmarks/src/radiantModels.test.ts
mongodben: i dont think we need to follow up b/c we only run the benchmarks from our local envs, so don't need to make sure it works from the cloud. this is just to smoke test that everything is functional w/ radiant

PR #463, opened 2024-08-07
Add analytics views on scrubbed_messages

nlarew:
## Notes

- For use in reporting dashboard - rather than duplicating aggregation pipelines in the Charts UI we can pull from these views.

### DIFF
--- a/packages/scripts/src/createView/scrubbed_messages_stats/daily.ts
+++ b/packages/scripts/src/createView/scrubbed_messages_stats/daily.ts
@@ -1,0 +1,95 @@
+/**
+  Creates views with aggregated analytics stats. This is useful for things
+  like our dashboard which charts historical usage.
+ */
+
+import "dotenv/config";
+import { Db, Document } from "mongodb-rag-core";
+import { ChatbotStats } from ".";
+
+export type ScrubbedMessageStatsForDay = Document &
+  ChatbotStats & {
+    date: Date;
+    year: number;
+    dayOfYear: number;
+    month: number;
+    dayOfMonth: number;
+    week: number;
+    dayOfWeek: number;
+  };
+
+export function createScrubbedMessageStatsForDayView({
+  db,
+  viewName,
+}: {
+  db: Db;
+  viewName: string;
+}) {
+  return db.createCollection<ScrubbedMessageStatsForDay>(viewName, {
+    viewOn: "scrubbed_messages",
+    pipeline: [
+      {
+        $match: {
+          role: "user",
+        },
+      },
+      {
+        $group: {
+          _id: {
+            dayOfYear: {
+              $dayOfYear: "$createdAt",
+            },
+            isoYear: {
+              $isoWeekYear: "$createdAt",
+            },
+          },
+          numMessages: {
+            $sum: 1,
+          },
+          conversations: {
+            $addToSet: "$conversationId",
+          },
+        },
+      },
+      {
+        $project: {
+          _id: 0,
+          year: "$_id.isoYear",
+          dayOfYear: "$_id.dayOfYear",
+          date: {
+            $dateFromParts: {
+              year: "$_id.isoYear",
+              day: "$_id.dayOfYear",
+            },
+          },
+          numMessages: 1,
+          numConversations: {
+            $size: "$conversations",
+          },
+        },
+      },
+      {
+        $addFields: {
+          month: {
+            $month: "$date",
+          },
+          dayOfMonth: {
+            $dayOfMonth: "$date",
+          },
+          week: {
+            $week: "$date",
+          },
+          dayOfWeek: {
+            $dayOfWeek: "$date",
+          },
+        },
+      },
+      {
+        $sort: {
+          year: -1,
+          dayOfYear: -1,
+        },
+      },
+    ],
+  });
+}
--- a/packages/scripts/src/createView/scrubbed_messages_stats/index.ts
+++ b/packages/scripts/src/createView/scrubbed_messages_stats/index.ts
@@ -1,0 +1,42 @@
+import { assertEnvVars, MongoClient } from "mongodb-rag-core";
+import { createScrubbedMessageStatsForDayView } from "./daily";
+import { createScrubbedMessageStatsForWeekView } from "./weekly";
+import { createScrubbedMessageStatsForMonthView } from "./monthly";
+
+export type ChatbotStats = {
+  numMessages: number;
+  numConversations: number;
+};
+
+const { MONGODB_DATABASE_NAME, MONGODB_CONNECTION_URI } = assertEnvVars({
+  MONGODB_DATABASE_NAME: "",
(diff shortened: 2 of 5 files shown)


packages/scripts/src/createView/scrubbed_messages_stats/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+    await db.dropCollection(monthlyViewName);
+    await createScrubbedMessageStatsForMonthView({
+      db,
+      viewName: monthlyViewName,
+    });
+  } finally {
+    await client.close();
+  }
+})();
```
mongodben, line 42: nit: can it be the function definition and then call afterwards instead of IFFE to follow pattern of other projects in here

packages/scripts/src/createView/scrubbed_messages_stats/daily.ts
```diff
@@ -0,0 +1,95 @@
...
+    date: Date;
+    year: number;
+    dayOfYear: number;
+    month: number;
+    dayOfMonth: number;
+    week: number;
+    dayOfWeek: number;
+  };
+
+export function createScrubbedMessageStatsForDayView({
```
mongodben, line 21: rather than 3 separate files each w separate function, can we combine all the these functions as methods under 1 interface to reduce duplication of code

mongodben (changes requested):
see comments and please add testing.

packages/scripts/src/createView/scrubbed_messages_stats/daily.ts
nlarew: Updated to use a "`makeCreator()`" pattern

packages/scripts/src/createView/scrubbed_messages_stats/index.ts
nlarew: Sure, updated

nlarew:
We don't really have anything set up to test view creation. Ultimately it's not mission critical if these break - we'll just notice that our charts aren't working.

nlarew:
@mongodben I modified this to be a _materialized view_ instead of an on-demand view. i.e. this is now an aggregation pipeline ending with a `$merge` that we run from a cron job. I think this will be a good base for any type of aggregated analytics we do in the future.

This shaves off ~50 seconds (currently, more over time) of aggregation query time any time we load/refresh/modify a chart. Basically we'll trade off a bit of software complexity & data that's up to 2 hours old in exchange for tons of time and compute.

packages/scripts/environments/production.yml
```diff
@@ -65,6 +65,22 @@ cronJobs:
         memory: 5Gi
     backoffLimit: 3
 
+  - name: materialize-scrubbed-message-stats
+    schedule: "0 */2 * * *" # Every 2 hours
+    command: ["npm", "run", "scripts:materializeScrubbedMessagesStats:latest"]
+    env:
+      MONGODB_DATABASE_NAME: docs-chatbot-staging
+    envSecrets:
+      MONGODB_CONNECTION_URI: docs-chatbot-staging
```
mongodben, line 74: why is there a production script that points to our staging DB?

packages/scripts/src/createView/scrubbed_messages_stats.ts
mongodben, line 1: nit on file name: why snake_case? should be camelCase, no?

packages/scripts/src/materialize/materialized-view-utils.ts
mongodben, line 1: nit on file name: why kebab-case? should be camelCase, no?

mongodben:
> Happy to walk through this during standup, probably easier to just point everything out.

yes pls. maybe we can today?

packages/scripts/environments/production.yml
nlarew: Because I originally copied this in the staging env file and forgot to update 🤦🏻 Will fix!

packages/scripts/src/createView/scrubbed_messages_stats.ts
nlarew: (I actually meant to delete this particular file for the on-demand view - still applies to the materialized view file)

Because it matches the view/collection name in MongoDB - we do the same thing for the other `createView/*` files. Doesn't really matter to me, we could do snake case here.

packages/scripts/environments/production.yml
mongodben, line 1: should this also run on our staging DB?

packages/scripts/src/materialize/materializedViewUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+  db: Db;
+  collectionName: string;
+} & ({ index: IndexDefinition } | { indexes: IndexDefinition[] });
+
+export async function ensureCollectionWithIndex({
+  db,
+  collectionName,
+  ...indexProps
+}: EnsureCollectionWithIndexParams) {
+  console.log("Ensuring collection with index:", {
```
mongodben, line 24: should be `logger.info()` i think to be consistent w our other stuff

mongodben: ditto below `console.log`s

packages/scripts/src/materialize/materializedViewUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+// Date utils
+export function startOfMonth(date: Date) {
+  return new Date(Date.UTC(date.getFullYear(), date.getMonth(), 1, 0, 0, 0, 0));
+}
+
+export function startOfWeek(
+  date: Date,
+  weekStartsOn: "sunday" | "monday" = "sunday"
+) {
+  // The week starts on Monday
```
mongodben, line 79: can remove?

```suggestion
```

packages/scripts/src/materialize/materializedViewUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Date utils
+export function startOfMonth(date: Date) {
+  return new Date(Date.UTC(date.getFullYear(), date.getMonth(), 1, 0, 0, 0, 0));
+}
+
+export function startOfWeek(
+  date: Date,
+  weekStartsOn: "sunday" | "monday" = "sunday"
```
mongodben, line 77: does this need to be an option? feels goldplatey

mongodben (approved):
lgtm % few things.

also note tests failing https://drone.corp.mongodb.com/mongodb/chatbot/3267/1/2. should be addressed before merge

packages/scripts/environments/production.yml
nlarew: Didn't put it there initially since we don't really need stats, but I'll add it for consistency

packages/scripts/src/materialize/materializedViewUtils.ts
nlarew: We don't have a logger library for the `scripts` directory, lots of `console.log()` in this package

PR #472, opened 2024-08-15
(EAI-481) Clean up env vars

nlarew:
Jira: (EAI-481) Clean up env vars

## Changes

- Updates to our env vars so that each project only asks for / requires the vars it actually needs and uses.
  - Updates `.env.example` files
  - Breaks `CORE_ENV_VARS` into sub-objects that are more granular + adds similar objects for the ingest packages

- Removes `OPENAI_EMBEDDING_MODEL` and `OPENAI_EMBEDDING_MODEL_VERSION` from most projects
  - Removed from k8s envs (`environments/{staging|production}.yml`)
  - We aren't actually using these values in many cases as they're redundant with the model deployment name.
  - These vars are still in `mongodb-chatbot-verified-answers` because it stores them in the db - we should probably re-evalaute our strategy here.


packages/ingest-mongodb-public/.env.example
```diff
@@ -1,26 +1,16 @@
...
-MONGODB_CONNECTION_URI=<connection_uri>
-MONGODB_DATABASE_NAME=<docs-chatbot-ENV>
-VECTOR_SEARCH_INDEX_NAME=<index_name>
-NODE_ENV=development
-# Connection string for DevCenter
-DEVCENTER_CONNECTION_URI="mongodb+srv://<username>:<password>@devhub-cluster.sewho.mongodb.net/?retryWrites=true&w=majority"
-# For connecting to MongoDB University Data API
-UNIVERSITY_DATA_API_KEY=abc123
-# For use with GitHubDataSource. The token only needs read access to the repo.
-GITHUB_ACCESS_TOKEN=ghp_TOKENTOKENTOKEN
```
mongodben, line 7:
```
# Connection string for DevCenter
DEVCENTER_CONNECTION_URI="mongodb+srv://<username>:<password>@devhub-cluster.sewho.mongodb.net/?retryWrites=true&w=majority"
# For connecting to MongoDB University Data API
```

i think this is needed

packages/ingest-mongodb-public/src/PublicIngestEnvVars.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  CORE_CHATBOT_APP_ENV_VARS,
+  CORE_OPENAI_EMBEDDING_ENV_VARS,
+} from "mongodb-rag-core";
+import { INGEST_MDBU_ENV_VARS } from "mongodb-rag-ingest";
+
```
mongodben, line 6: also need dev center DB i think

packages/mongodb-chatbot-server/README.md
```diff
@@ -55,7 +55,6 @@ export const {
   OPENAI_ENDPOINT,
   OPENAI_API_KEY,
   OPENAI_EMBEDDING_DEPLOYMENT,
-  OPENAI_EMBEDDING_MODEL_VERSION,
   OPENAI_CHAT_COMPLETION_MODEL_VERSION,
```
mongodben, line 58: can delete this too?

packages/mongodb-chatbot-server/src/test/testConfig.ts
```diff
@@ -38,7 +38,6 @@ export const {
   OPENAI_ENDPOINT,
   OPENAI_API_KEY,
   OPENAI_EMBEDDING_DEPLOYMENT,
-  OPENAI_EMBEDDING_MODEL_VERSION,
   OPENAI_CHAT_COMPLETION_MODEL_VERSION,
```
mongodben, line 41: can delete this too?

packages/mongodb-chatbot-verified-answers/.env.example
```diff
@@ -1,6 +1,6 @@
 MONGODB_DATABASE_NAME=<docs-chatbot-ENV>
 MONGODB_CONNECTION_URI=<connection_uri>
-OPENAI_EMBEDDING_MODEL=text-embedding-ada-002
+OPENAI_EMBEDDING_MODEL_NAME=text-embedding-ada-002
```
mongodben, line 3: can this just be removed?

packages/mongodb-chatbot-verified-answers/.env.example
```diff
@@ -1,6 +1,6 @@
...
 OPENAI_EMBEDDING_MODEL_VERSION=2023-03-15-preview
```
mongodben, line 4: i think can also remove this

packages/mongodb-chatbot-verified-answers/.env.example
mongodben: rather, we should use "OPENAI_EMBEDDING_DEPLOYMENT" for consistency w/ other packages

mongodben:
> These vars are still in mongodb-chatbot-verified-answers because it stores them in the db - we should probably re-evalaute our strategy here.

can you make ticket for this?

packages/ingest-mongodb-public/.env.example
nlarew: Are you sure? All of the tests pass without it + we make no reference to it in `ingest-mongodb-public` (although we probably should move _all_ of the devcenter/mdbu stuff into our mongodb ingest package instead of core ingest)

mongodben: > we probably should move all of the devcenter/mdbu stuff into our mongodb ingest package instead of core ingest

yeah, and the snooty stuff too

mongodben: yeah, it seems we import it directly from the env in the constructor (bad!) https://github.com/mongodb/chatbot/blob/main/packages/mongodb-rag-ingest/src/sources/DevCenterDataSource.ts#L38

so for now, we should keep this env var here. and maybe we could have a follow up ticket to pass this as a property to the `makeDevCenterDataSource()` constructor.

packages/ingest-mongodb-public/src/PublicIngestEnvVars.ts
nlarew: Updated the import to use `INGEST_ENV_VARS` (which includes dev center and MDBU) instead of just `INGEST_MDBU_ENV_VARS`

packages/mongodb-chatbot-server/README.md
nlarew: Not yet, we still reference it in our analyzer code. Ticket to update that https://jira.mongodb.org/browse/EAI-502

packages/mongodb-chatbot-verified-answers/.env.example
nlarew: https://jira.mongodb.org/browse/EAI-502

packages/mongodb-chatbot-verified-answers/.env.example
nlarew: https://jira.mongodb.org/browse/EAI-502

PR #478, opened 2024-08-20
(EAI-484) Fix broken langchain chat model wrapper

nlarew:
Jira: (EAI-484) Fix broken langchain chat model wrapper

## Changes

- Upgrade to the latest `@langchain/*` dependencies.
  - Core wasn't actually referencing `@langchain/anthropic` as a dependency, so this adds that.
- Removes the `as unknown as` casts from https://github.com/mongodb/chatbot/pull/465#discussion_r1718834261

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41332,7 +41332,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@azure/openai": "^1.0.0-beta.11",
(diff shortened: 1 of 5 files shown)


mongodben (approved):
lgtm, can you do follow up ticket(s) to publish the packages

PR #486, opened 2024-08-26
Update quick start guide documentation

Rebecca-Qian:
## Changes

- Fix typo in quickstart guide to point to `http://localhost:3000/api/v1/conversations/`
- Add instructions for appending new messages to the conversation

## Notes

- Example here uses `stream=false`, but users may want to use `stream=true` for longer responses
- Should note that the [default max input](https://github.com/mongodb/chatbot/blob/844bbf5e83ee8a7b6be1077c3b2407543429c0fb/packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts#L40) is 300 chars, which limits the types of inputs users can provide. Temporary workaround is to override by setting `maxInputLengthCharacters` in the example [config](https://github.com/mongodb/chatbot/blob/main/examples/quick-start/packages/server/src/index.ts#L125)

### DIFF
--- a/docs/docs/quick-start.md
+++ b/docs/docs/quick-start.md
@@ -172,10 +172,21 @@ You can ask the chatbot questions and see the responses.
 
 Have fun!
 
-You can also query the server directly with curl:
+You can also query the server directly with curl. To create a new conversation:
 
 ```shell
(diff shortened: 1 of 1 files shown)


mongodben (approved):
thanks!

PR #493, opened 2024-09-04
(EAI-495) [UI] Support references on ChatWindow initial message

nlarew:
Jira: (EAI-495) [UI] Support references on ChatWindow initial message

## Changes

- Adds a new optional `initialMessageReferences` prop to define reference links on the initial message
- Adds a check in `Message.tsx` to only render the references section if at least one reference is passed (i.e. handle the empty array case)

Screenshot 2024-09-04 at 12 18 20 PM


mongodben (approved):
lgtm, should have docs too right?

PR #489, opened 2024-08-28
(EAI-506): Refactor full response generation as stateless function

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-506

## Changes

- Refactor the response generation as a stateless function so that we can more easily evaluate outputs w/o a server
- Evaluate our whole RAG pipeline w/ braintrust autoevals
- Refactor autoevals to use azure openai instead of plain ol openai
- Upgrade typechat to `"^0.1.0"` b/c it was causing issues w/ braintrust CLI (see https://github.com/braintrustdata/braintrust-sdk/issues/356)
- Add Braintrust instrumentation to our server implementation to trace results in the evals.
  - Note: We can also easily trace w the braintrust observability module if ever choose to use that.
- Refactor server unit tests for simplicity

## Notes

- Example Braintrust run: https://www.braintrust.dev/app/mdb-test/p/mongodb-chatbot-conversations/experiments/mongodb-chatbot-latest

PR #498, opened 2024-09-06
(EAI-506 redux): Refactor full response generation as a stateless function

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-506

## Changes

- Refactor the response generation as a stateless function so that we can more easily evaluate outputs w/o a server
- Evaluate our whole RAG pipeline w/ braintrust autoevals
- Refactor autoevals to use azure openai instead of plain ol openai
- Upgrade typechat to `"^0.1.0"` b/c it was causing issues w/ braintrust CLI (see https://github.com/braintrustdata/braintrust-sdk/issues/356)
- Add Braintrust instrumentation to our server implementation to trace results in the evals.
  - Note: We can also easily trace w the braintrust observability module if ever choose to use that.
- Refactor server unit tests for simplicity

(Taken from #489)

## Notes

- Revised version of #489. Reverted in #497 due to issues caused in staging, see that PR description for docs on issues observed.


packages/mongodb-chatbot-server/src/routes/generateResponse.test.ts
```diff
@@ -231,6 +231,36 @@ describe("generateResponse", () => {
...
+            content: "test metadata",
+          },
+          staticResponse,
+        };
+      },
+    });
+
+    const data = res._getData();
+
+    // The metadata should be streamed before the assistant response
```
nlarew, line 258: nit - I'm not sure that this is necessarily a guarantee.

For example, we in theory support the following:
1. half of the assistant message gets streamed in
2. A `metadata` event arrives that indicates the message is verified (or some other future metadata-driven state)
3. The UI updates to reflect the latest metadata
4. The rest of the assistant message streams in

nlarew (approved):

packages/mongodb-chatbot-server/src/routes/generateResponse.test.ts
```diff
@@ -443,4 +529,18 @@ describe("streamGenerateResponse", () => {
...
   });
+  it("should stream metadata", async () => {
+    const metadata = { foo: "bar", baz: 42 };
+    await streamGenerateResponseMessage({
+      ...baseArgs,
+      metadata,
+    });
+    const data = res._getData();
+
+    // The metadata should be streamed before the assistant response
```
mongodben, line 540:
```suggestion
```

PR #505, opened 2024-09-17
(EAI-534): Script to load benchmark sample data

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-534

## Changes

- Script to create sample databases in MongoDB. Loads data from Braintrust.
- Created new `sample-data` cluster for this data. 

## Notes

-


packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
```diff
@@ -0,0 +1,108 @@
...
+  apiKey,
+  projectName,
+  datasetName = "database-documents",
+}: LoadBraintrustDatasetParams) {
+  const dataset = initDataset(projectName, {
+    apiKey,
+    dataset: datasetName,
+  });
+  const dbDocuments: DbDocument[] =
+    // TODO: why do i have to cast this as unknown?
```
nlarew, line 81: I pulled the PR down and don't get a type error if I remove the unknown, i.e this builds successfully for me:

```ts
((await dataset.fetchedData()) as DatabaseMetadataDatasetEntry[]).map(...)
```

Still sucks that we need the type cast. I think it might be solved with some better generics in the braintrust lib. In the meantime, we could avoid this cast altogether by parsing with `zod` instead.

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
```diff
@@ -0,0 +1,108 @@
...
+  collections: CollectionMetadata[];
+}
+
+export interface CollectionMetadata {
+  collectionName: string;
+  indexes: {
+    v: CreateIndexesOptions["version"];
+    key: Map<string, IndexDirection>;
+    name: CreateIndexesOptions["name"];
+    // collation?: CollationOptions;
```
nlarew, line 67: Do we need this? If not can we remove it + its unused import?

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
```diff
@@ -0,0 +1,108 @@
...
+      (await dataset.fetchedData()) as unknown as DatabaseMetadataDatasetEntry[]
+    ).map(
+      (m) =>
+        ({
+          databaseName: m.input.databaseName,
+          collections: m.input.collections.map((c) => ({
+            collectionName: c.collectionName,
+            indexes: c.indexes.map((idx) =>
+              EJSON.deserialize(idx)
+            ) as unknown as CollectionMetadata["indexes"],
```
nlarew, line 38: Don't love this cast either, might be solved with zod parsing?

nlarew (approved):
LGTM overall, left a few small suggestions

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
```diff
@@ -0,0 +1,108 @@
+import { initDataset } from "braintrust";
+import { EJSON } from "bson";
```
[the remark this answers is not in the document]
mongodben: > Rather than using bson as a standalone library, can we use the version bundled with mongodb/exported from mongodb-rag-core?

i feel like i tried this and it didn't work but can take another look

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
mongodben: >  I think it might be solved with some better generics in the braintrust lib

agree. they're generally pretty good about supporting generics. surprised they don't have it for this method. will 2x check to make sure that the generic isn't supported and file a ticket for them if it in fact is not

> we could avoid this cast altogether by parsing with zod instead.

yea, good point. let me throw that together

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
mongodben: yes, can remove. good catch

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
mongodben: i think so. will implement

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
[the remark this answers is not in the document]
mongodben: yep this worked

packages/benchmarks/src/textToDriver/makeDatabases/loadBraintrustDatasets.ts
mongodben: the removing cast as unknown is also working for me..not sure what was up before. was getting funky ts err. 

anyways, converted to zod types and using that. 

aside: gh copilot is magic at taking a TS type and creating the zod type for it.

mongodben: created issue for braintrust https://github.com/braintrustdata/braintrust-sdk/issues/384

PR #508, opened 2024-09-18
(EAI-537): Execute generated Node.js driver code from string

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-537

## Changes

- Function to execute generated Node.js driver code from a string using `eval()`
-

## Notes

-


packages/benchmarks/src/textToDriver/executeGeneratedDriverCode.ts
```diff
@@ -0,0 +1,83 @@
...
+  // While the database variable looks unused, it is actually used in the generated code
+  const database = mongoClient.db(databaseName);
+
+  const startTime = Date.now();
+
+  try {
+    // Wrap the generated code in an async IIFE to guarantee that it returns a Promise
+    const result = await eval(
+      `(async () => { return ${generatedDriverCode} })()`
+    );
```
nlarew, line 65: We should consider using something more secure/controllable than a raw `eval`. Maybe node's built-in `vm` package or something with stronger security guarantees like https://www.npmjs.com/package/isolated-vm.

Quick read: https://odino.org/eval-no-more-understanding-vm-vm2-nodejs/

I could see some type of attack vector where you take a well known query (e.g. from a public benchmark dataset) and seed LLMs (e.g. via scraping sites you create for this purpose) with a malicious JS payload.

Maybe not a huge risk, but something to consider!

mongodben: > We should consider using something more secure/controllable than a raw eval.

i considered using the `vm` module, but it seemed that there could be issues passing the mongodb database object, which is using the MongoClient instance to a separate process. but i don't think i actually tried it out. will give it a crack and we can use if its not an issue

mongodben: update: vm.runInNewContext() seems to work. see updated code

packages/benchmarks/src/textToDriver/executeGeneratedDriverCode.ts
```diff
@@ -0,0 +1,84 @@
...
+): Promise<{
+  result: Document | Document[] | number | null;
+  error?: {
+    message: string;
+  };
+  executionTimeMs: number;
+}> {
+  const { generatedDriverCode, mongoClient, databaseName } = params;
+  // While the database variable looks unused, it is actually used in the generated code
+  const database = mongoClient.db(databaseName);
```
nlarew, line 58:
```suggestion
  const database = mongoClient.db(databaseName);
```
Now it's actually used with `vm` :)

packages/benchmarks/src/textToDriver/executeGeneratedDriverCode.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const startTime = Date.now();
+
+  // Wrap the generated code in an async IIFE to guarantee that it returns a Promise
+  const promiseCode = `(async () => { return ${generatedDriverCode} })()`;
```
nlarew, line 63:
```suggestion
  // Wrap the generated code in an async IIFE to guarantee that it returns a Promise
  const promiseCode = `(async () => { return ${generatedDriverCode} })()`;
  
  const startTime = Date.now();
```

Very likely negligible _but_ I think `startTime` should technically go after the `promiseCode` declaration since that's setup and not the actual code execution we want to time.

nlarew (approved):
LGTM w/ a couple nitpicks

PR #509, opened 2024-09-19
(EAI-418): use concurrency when ingesting and embedding

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-418

## Changes

- Using promise-pool to process tasks relating to ingesting and embedding concurrently rather than sequentially


packages/mongodb-rag-ingest/src/Config.ts
```diff
@@ -45,6 +45,34 @@ export type Config = {
...
+  /**
+    Options for concurrency.
+   */
+  concurrencyOptions?: Constructor<ConcurrencyOptions>;
 };
 
+
+  /**
+    Options for concurrency. 
+    Set the number of tasks to process concurrently in the promise pool.
```
mongodben, line 58:
```suggestion
    Set the number of concurrent promises to execute for the given tasks.
```

packages/mongodb-rag-ingest/src/Config.ts
```diff
@@ -45,6 +45,34 @@ export type Config = {
...
+    If not specified, tasks will be run sequentially.
+   */
+export interface ConcurrencyOptions {
+  /**
+    Options for concurrency when embedding content. Set the number of tasks to process concurrently.
+   */
```
mongodben, line 64:
```suggestion
  /**
    Options for concurrency when embedding content with the `embed` command.
   */
```

packages/mongodb-rag-ingest/src/Config.ts
```diff
@@ -45,6 +45,34 @@ export type Config = {
...
+  embed?: {
+    processPages: number,
+    createChunks: number,
+  },
+  /**
+    Options for concurrency for pages. Set the number of tasks to process concurrently.
+   */
```
mongodben, line 71:
```suggestion
  /**
    Options for concurrency when ingesting
    with the `pages` command.
   */
```

packages/mongodb-rag-ingest/src/Config.ts
```diff
@@ -45,6 +45,34 @@ export type Config = {
...
+  pages?: {
+    processDataSources: number,
```
mongodben, line 73:
```suggestion
    /**
      Number of data sources to process concurrently.
     */
    processDataSources?: number,
```

packages/mongodb-rag-ingest/src/commands/pages.ts
```diff
@@ -53,5 +54,6 @@ export const doPagesCommand = async (
   await updatePages({
     sources,
     pageStore,
+    concurrencyOptions
```
mongodben, line 57: rather than passing the whole `concurrencyOptions` object, you should just pass concurrencyOptions.pages`

mongodben: same comment applies for embedding

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.ts
```diff
@@ -127,25 +136,25 @@ export const updateEmbeddedContentForPage = async ({
...
+      const { embedding } = await embedder.embed({
+        text: chunk.text,
+      });
+      embeddedContent.push({
+        ...chunk,
+        embedding,
+        updated: new Date(),
+        chunkAlgoHash,
+      });
+    })
```
mongodben, line 157: rather than pushing into the `embeddedContent` array, you can just return the results from the PromisePool. yields cleaner code and takes advantage of any optimizations in the PromisePool library.

something like:

```ts
const embeddedContent = await PromisePool...process((args)=>{
  // ...
  return {
    // the embedded content object
  }
})
```

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
mongodben, line 1: upon further thought, we should have some form of basic tests for this. we should not leave configuration options untested. 

even if we trust the promisepool implementation works, we still want to test to make sure that our API is doing what it's supposed to

packages/mongodb-rag-ingest/src/pages/updatePages.ts
mongodben, line 1: need tests for this too

mongodben (changes requested):
nice work, i think this is mostly there. see comments.

regarding the testing aspect, we can do a quick pair on that to figure out the approach to take. pls schedule some time for this.

packages/mongodb-rag-ingest/src/Config.ts
```diff
@@ -45,6 +45,46 @@ export type Config = {
...
+    Options for concurrency when ingesting
+    with the `pages` command.
+   */
+  pages?: {
+    /**
+      Number of data sources to process concurrently.
+     */
+    processDataSources?: number,
+  }
+}
```
mongodben, line 87: pls add line breaks before all doc comments like below. 

chris had a preference for it. and it's a convention we try to uphold

```suggestion
export interface ConcurrencyOptions {

  /**
    Options for concurrency when chunking and embedding content
    with the `embed` command.
   */
  embed?: {
 
    /**
      Number of pages to chunk and embed concurrently.
     */
    processPages?: number,

    /**
      Maximum number of chunks per page to generate chunks for concurrently.
      This includes the creation of the chunk embeddings and any chunk preprocessing.
     */
    createChunks?: number,
  },

  /**
    Options for concurrency when ingesting
    with the `pages` command.
   */
  pages?: {

    /**
      Number of data sources to process concurrently.
     */
    processDataSources?: number,
  }
}
```

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.ts
```diff
@@ -20,13 +22,15 @@ export const updateEmbeddedContent = async ({
...
   chunkOptions,
+  embedConcurrencyOptions,
 }: {
   since: Date;
   embeddedContentStore: EmbeddedContentStore;
   pageStore: PageStore;
   embedder: Embedder;
   chunkOptions?: Partial<ChunkOptions>;
   sourceNames?: string[];
+  embedConcurrencyOptions?: ConcurrencyOptions["embed"];
```
mongodben, line 33: style: please call this property just `concurrencyOptions`. the fact that it's on the function `updateEmbeddedContent()`, makes the *embed* part implicitly redundant


```suggestion
  concurrencyOptions?: ConcurrencyOptions["embed"];
```

mongodben: also, i think better interface design would be to define `EmbedConcurrencyOptions` in this file, ie:

```ts
export interface EmbedConcurrencyOptions {
 
    /**
      Number of pages to chunk and embed concurrently.
     */
    processPages?: number,
    /**
      Maximum number of chunks per page to generate chunks for concurrently.
      This includes the creation of the chunk embeddings and any chunk preprocessing.
     */
    createChunks?: number,
}
```

and then use that here, such as:

```suggestion
  concurrencyOptions?: EmbedConcurrencyOptions;
```

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.ts
```diff
@@ -37,26 +41,29 @@ export const updateEmbeddedContent = async ({
...
-        await updateEmbeddedContentForPage({
-          store: embeddedContentStore,
-          page,
-          chunkOptions,
-          embedder,
-        });
-    }
-  }
+  await PromisePool
+    .withConcurrency(embedConcurrencyOptions?.processPages || 1)
```
mongodben, line 45:
```suggestion
    .withConcurrency(embedConcurrencyOptions?.processPages ?? 1)
```

prefer [nullish coalescing (`??`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) instead of or operating (`||`)

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.ts
```diff
@@ -125,27 +134,25 @@ export const updateEmbeddedContentForPage = async ({
...
-    });
-    embeddedContent.push({
-      ...chunk,
-      embedding,
-      updated: new Date(),
-      chunkAlgoHash,
-    });
-  }
+  const { results: embeddedContent } = await PromisePool
+    .withConcurrency(embedConcurrencyOptions?.createChunks || 1)
```
mongodben, line 138:
```suggestion
    .withConcurrency(embedConcurrencyOptions?.createChunks ?? 1)
```

packages/mongodb-rag-ingest/src/pages/updatePages.ts
```diff
@@ -1,36 +1,41 @@
...
   store.
  */
 export const updatePages = async ({
   sources,
   pageStore,
+  pageConcurrencyOptions,
 }: {
   sources: DataSource[];
   pageStore: PageStore;
+  pageConcurrencyOptions?: ConcurrencyOptions["pages"]
```
mongodben, line 17: same comments as for the `updateEmbeddedContent` about the parameter naming + typescript typing

packages/mongodb-rag-ingest/src/pages/updatePages.ts
```diff
@@ -1,36 +1,41 @@
...
-      continue;
-    }
-    await persistPages({
-      pages,
-      store: pageStore,
-      sourceName: source.name,
-    });
-  }
+  await PromisePool
+    .withConcurrency(pageConcurrencyOptions?.processDataSources || 1)
```
mongodben, line 20:
```suggestion
    .withConcurrency(pageConcurrencyOptions?.processDataSources ?? 1)
```

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -138,4 +146,64 @@ describe("updateEmbeddedContent", () => {
...
+
+    const since = new Date("2000-01-01");
+
+    const startTimes: number[] = [];
+    const endTimes: number[] = [];
+
+    const originalDateNow = Date.now;
+    let currentFakeTime = originalDateNow();
+
+    jest.spyOn(global.Date, 'now').mockImplementation(() => currentFakeTime);
```
mongodben, line 173: i don't think you need this

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -138,4 +146,64 @@ describe("updateEmbeddedContent", () => {
...
+
+    const mockEmbedder: jest.Mocked<Embedder> = {
+      embed: jest.fn().mockImplementation(async (param) => {
+        const startTime = Date.now();
+        startTimes.push(startTime);
+        await new Promise((resolve) => setTimeout(resolve, Math.random() * 100)); // Random delay to simulate processing
```
mongodben, line 179: why not just do a static amount, such as 50ms?

```suggestion
        await new Promise((resolve) => setTimeout(resolve, 50));
```

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -138,4 +146,64 @@ describe("updateEmbeddedContent", () => {
...
+        return { embedding: [1, 2, 3] };
+      }),
+    } as unknown as jest.Mocked<Embedder>;
+
+    await updateEmbeddedContent({
+      embedder: mockEmbedder,
+      embeddedContentStore,
+      pageStore,
+      since,
+      embedConcurrencyOptions: { processPages: 2, createChunks: 2 },
```
mongodben, line 192: i am not sure if you're testing both of these. 

anyways, i think these should be separate tests, one for `processPages` and another for `createChunks`

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -138,4 +146,64 @@ describe("updateEmbeddedContent", () => {
...
+    // Ensure some overlaps indicating concurrency
+    expect(executionPairs.some((pair, i, pairs) =>
+      pairs.some((otherPair, j) =>
+        i !== j &&
+        pair.startTime < otherPair.endTime &&
+        otherPair.startTime < pair.endTime
+      ))
+    ).toBe(true);
+  });
+});
```
mongodben, line 209:
```suggestion
});

```

add new line at EOF

packages/mongodb-rag-ingest/src/pages/updatePages.test.ts
```diff
@@ -66,3 +67,67 @@ describe("persistPages", () => {
...
+    
+    const sources = [source1, source2];
+    
+    const startTimes: number[] = [];
+    const endTimes: number[] = [];
+
+    const originalDateNow = Date.now;
+    let currentFakeTime = originalDateNow();
+
+    jest.spyOn(global.Date, 'now').mockImplementation(() => currentFakeTime);
```
mongodben, line 99: similar to above on the embedded content stuff, not sure if this all is necessary. think can just do the awaited promises

packages/mongodb-rag-ingest/src/pages/updatePages.test.ts
```diff
@@ -66,3 +67,67 @@ describe("persistPages", () => {
...
+    // Ensure some overlaps indicating concurrency
+    expect(executionPairs.some((pair, i, pairs) =>
+      pairs.some((otherPair, j) =>
+        i !== j &&
+        pair.startTime < otherPair.endTime &&
+        otherPair.startTime < pair.endTime
+      ))
+    ).toBe(true);
+  });
+});
```
mongodben, line 133:
```suggestion
});

```

mongodben (changes requested):
good work, few things throughout to address. then should be good to go

mongodben:
also, please update the PR title to follow our conventions of `(EAI-###): <some description>`

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -1,7 +1,17 @@
...
+  Page,
+  PersistedPage,
+} from "mongodb-rag-core";
+import {
+  updateEmbeddedContent,
+  updateEmbeddedContentForPage,
+} from "./updateEmbeddedContent";
 import { persistPages } from "../pages";
 import { makeMockPageStore } from "../test/MockPageStore";
+import * as chunkPageModule from "./chunkPage";
```
mongodben, line 14: can you just import the function `chunkPage()`, and spy on that rather than do a * import?

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -138,4 +148,139 @@ describe("updateEmbeddedContent", () => {
...
+  it("processes chunks concurrently within a page", async () => {
+    const embeddedContentStore = makeMockEmbeddedContentStore();
+    const page: PersistedPage = {
+      ...examplePage,
+      updated: new Date(),
+      action: "updated",
+    };
+
+    // Spy and mock chunkPage
+    const chunkPageSpy = jest
```
mongodben, line 160: as an aside: i think this is a reasonable approach b/c we don't have the `chunkPage()` function as something that you can pass as config...but we really should (perhaps keeping this as the default).

can you see if we have a ticket for this, and if not, create one?

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
```diff
@@ -138,4 +148,139 @@ describe("updateEmbeddedContent", () => {
...
+    });
+
+    const embeddedContentStore = makeMockEmbeddedContentStore();
+
+    const since = new Date("2000-01-01");
+
+    const startTimes: number[] = [];
+    const endTimes: number[] = [];
+
+    const mockEmbedder: jest.Mocked<Embedder> = {
```
mongodben, line 250: can you share this mock embedded with the other tests? 

better to not redefine per file

mongodben (changes requested):
looking better. just a few small things in the tests. lmk if you want to pair on those to resolve quickly.

packages/mongodb-rag-ingest/src/embed/updateEmbeddedContent.test.ts
yakubova92: I cant seem to do that without having to use `jest.spyOn(require('./chunkPage'), 'chunkPage')`
which causes this eslint error: Require statement not part of import statement. 
If i change the import to `import { chunkPage } from './chunkPage'; `, the mocking doesn't work correctly and the test fails.

mongodben: ok lets leave it as is then

mongodben (approved):
lgtm, nice work here

now that we have this implemented in the framework, could you make another ticket to use these config options in our implementation in the package `ingest-mongodb-public`?

yakubova92:
> now that we have this implemented in the framework, could you make another ticket to use these config options in our implementation in the package `ingest-mongodb-public`?

This is making me think that maybe we should be using concurrency as a default. The promise-pool package we're using defaults to 10 when no concurrency number is provided. We're setting the default to 1, meaning no concurrency, which is essentially requiring users to set concurrency options if they want to enable concurrency at all. Do you think it makes sense for us to choose values that would make sense for most cases for pages and chunks and set that as our default? Setting concurrencyOptions would allow users to overwrite that if needed, but ingesting and embedding steps would use concurrency out of the box.

PR #511, opened 2024-09-23
(EAI-538): Braintrust eval orchestration for text to node driver

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-538

## Changes

- Add simple eval metrics
- Add model provider to radiant model config
- Code to load and execute benchmark evals

## Notes

-


packages/benchmarks/src/textToDriver/loadBraintrustDatasets.ts
```diff
@@ -115,3 +126,23 @@ export async function loadBraintrustDbDocuments({
...
   return dbDocuments;
 }
+
+/**
+  Load the evaluation cases for the text-to-driver task from Braintrust.
+ */
+export async function loadBraintrustEvalCases({
+  apiKey,
+  projectName,
+  datasetName = "text-to-query-results",
```
nlarew, line 136:
```suggestion
  datasetName,
```

Wondering if we should remove this default in favor of always being explicit about the dataset name?

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
```diff
@@ -0,0 +1,167 @@
...
+  const timeout = 60000;
+
+  const prompts = NODE_JS_PROMPTS.systemPrompts;
+  const mongoClient = new MongoClient(MONGODB_TEXT_TO_DRIVER_CONNECTION_URI);
+
+  const SAMPLE_DOCUMENT_LIMIT = 2;
+
+  try {
+    await mongoClient.connect();
+    await sleep(500);
```
nlarew, line 34: Curious what the sleep is for. Is there a potential race condition here (e.g. on a very slow network)?

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+      }
+      return modelExperiments;
+    });
+    // Process models in parallel
+    await PromisePool.for(modelExperiments)
+      .withConcurrency(4)
+      .process(async (modelInfos) => {
+        await PromisePool.for(modelInfos)
+          .withConcurrency(1)
```
nlarew, line 53: - Where are these concurrency values coming from?
- Could we flatten this to just use one promise pool instead of two?
  - A pool with concurrency one will likely perform worse than a regular `for...of` loop (though also likely negligible relative to LLM latency)

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
```diff
@@ -0,0 +1,167 @@
...
+                    mongoClient,
+                    limit: SAMPLE_DOCUMENT_LIMIT,
+                  },
+                },
+              });
+            }
+          );
+      });
+
+    await Promise.allSettled(
```
nlarew, line 100: What's this for? Seems like it's an alternative version of the same eval we just ran (calls `runTextToDriverEval` again with the same arguments)

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
mongodben: yeah, i was getting mongo errs w/o that, idk why..too many promises

packages/benchmarks/src/textToDriver/loadBraintrustDatasets.ts
mongodben: reasonable i think

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
mongodben: no, we want to run all the models in parallel while running 1 experiment at a time per model.

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
mongodben: i agree. this is probably silliness on my part. will take a good look and remove if not needed or otherwise explain.

i think this is an old worse version of the above

packages/benchmarks/src/textToDriver/evalTypes.ts
```diff
@@ -0,0 +1,92 @@
...
+  TextToDriverMetadata
+>;
+
+export const TextToDriverEvalCaseSchema: z.ZodSchema<
+  EvalCase<TextToDriverInput, TextToDriverExpected, TextToDriverMetadata>
+> = z.object({
+  input: TextToDriverInputSchema,
+  expected: TextToDriverExpectedSchema,
+  metadata: TextToDriverMetadataSchema,
+  // tags: z.array(z.string()).optional().nullable(),
```
nlarew, line 86:
```suggestion
```

Can delete this?

packages/benchmarks/src/textToDriver/evaluationMetrics.ts
```diff
@@ -0,0 +1,41 @@
...
+        }
+      : {}),
+  };
+};
+
+/**
+  Measure how long the query takes to execute in minutes.
+
+  Note: Measuring in minutes because
+  Braintrust throws an error if the score > 1.
```
nlarew, line 30: Not sure I totally follow this comment. Is the idea that we expect all execution times to be under 1 minute? Might be true now but not in the future, e.g. for large aggregations

nlarew (approved):
LGTM

packages/benchmarks/src/textToDriver/evaluationMetrics.ts
mongodben: yea exactly. this is limited in that respect

PR #513, opened 2024-09-23
(EAI-497) DevCenter reference links should be labeled Article

nlarew:
Jira: (EAI-497) DevCenter reference links should be labeled Article

## Changes

- Switch from Blog to Article
- Use variants directly from LG chat instead of hardcoding


packages/mongodb-chatbot-ui/src/messageLinks.tsx
```diff
@@ -42,3 +26,12 @@ export function formatReferences(
     return richLinkProps;
   });
 }
+
+export function getMessageLinks(
```
mongodben, line 30: should there be tests for this?

packages/chatbot-server-mongodb-public/src/processors/makeMongoDbReferences.ts
```diff
@@ -75,6 +75,7 @@ export function mongodbReferenceType(
       case "practical-aggregations-book":
         return "Book";
       case "devcenter":
+        return "Article";
```
mongodben, line 78: relevant tests now failing b/c of this change. pls update test to match

mongodben (approved):
lgtm % changes noted in review

packages/mongodb-chatbot-ui/src/messageLinks.tsx
nlarew: Yeah - I actually have some tests for this file in another PR. I'll port them over to this one

packages/chatbot-server-mongodb-public/src/processors/makeMongoDbReferences.ts
nlarew: Updated

PR #514, opened 2024-09-25
(EAI-536): Text-to-driver evaluation metrics

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-536

## Changes

- Create MQL to Spider-SQL fuzzy matcher
- Ad fuzzy matcher to eval metrics 

## Notes

-


packages/benchmarks/src/textToDriver/fuzzyMatch.test.ts
```diff
@@ -0,0 +1,403 @@
...
+    const result = fuzzyMatch({
+      mongoDbOutput,
+      expected,
+      isAggregation: true,
+    });
+
+    expect(result).toBe(true);
+  });
+
+  test("should return true when aggregation output has mismatched keys", () => {
```
nlarew, line 335: This is an interesting case! Is this the idea behind the `isAggregation` flag in general?

i.e. "We know the result will be `10`. Ideally it's in the `total` field but we actually don't _really_ care what the field name is for fuzzy matching."

nlarew (approved):

packages/benchmarks/src/textToDriver/fuzzyMatch.test.ts
mongodben: yeah exactly

PR #517, opened 2024-09-26
(EAI-480) Site-specific sort order for related resources

nlarew:
Jira: (EAI-480) Site-specific sort order for related resources

## Changes

- Adds a generic sort function factory, `makePrioritizeReferenceDomain()`, which accepts one or more URLs to prioritize
- Adds a MongoDB-specific sort function factory, `makePrioritizeCurrentMongoDbReferenceDomain()`, that prioritizes links to the current MongoDB domain
- Adds a prop, `Chatbot.sortMessageReferences`, for users of the library to define their own custom reference sort functions

## Notes

- This will work automatically on MongoDB domains. MongoDB use cases do not need to add any props.


packages/mongodb-chatbot-ui/src/sortReferences.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      "https://www.example123.com",
+    ]);
+  });
+});
+
+describe("normalizedHostname", () => {
+  it("removes the www. prefix from a URL's hostname", () => {
+    const url = new URL("https://www.mongodb.com");
+    expect(normalizedHostname(url)).toBe("mongodb.com");
+  });
```
mongodben, line 108: why include this functionality?

mongodben, line 108: nit: not sure if you want to treat `www` as always equal to base domain. while it's convention that they point to the same point, it's not required

packages/mongodb-chatbot-ui/src/sortReferences.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    // Neither URL matches any priority domain
+    return 0;
+  };
+}
+
+/**
+ * Get the hostname of a URL, normalized to omit the "www." prefix if it is present
+ */
+export function normalizedHostname(url: URL): string {
```
mongodben, line 51: nit cont: per prev comment this is the functionality i'm not sure should be in there.

maybe have as an optional flag somewhere instead of default...but that feels goldplate-y

packages/mongodb-chatbot-ui/src/messageLinks.ts
```diff
@@ -35,3 +40,27 @@ export function getMessageLinks(
...
 }
+
+export function makePrioritizeCurrentMongoDbReferenceDomain(): SortReferences {
+  const currentDomain = getCurrentPageUrl();
+  if (!currentDomain) {
+    // If we can't determine the current domain (e.g. on the server) then the sort is a no-op
+    return () => 0;
+  }
+  const prioritizableDomains = [
+    new URL("https://mongodb.com/docs"),
```
mongodben, line 51: aside: i always find the `URL` API quite confusing to work with and need to look up the API ref every time i see it

mongodben (approved):
LGTM. i like the implementation API. 

see one suggestion about `www`

packages/mongodb-chatbot-ui/src/sortReferences.ts
nlarew: Yeah I think it's unnecessary in hindsight

PR #529, opened 2024-10-02
(EAI-532) [UI] Fix spacing for headings and lists

nlarew:
Jira: (EAI-532) [UI] Fix spacing for headings and lists
Screenshot 2024-10-02 at 5 45 27 PM
Screenshot 2024-10-02 at 5 45 37 PM
Screenshot 2024-10-02 at 5 45 46 PM

### DIFF
--- a/packages/mongodb-chatbot-ui/src/Message.tsx
+++ b/packages/mongodb-chatbot-ui/src/Message.tsx
@@ -74,42 +74,23 @@ const styles = {
     display: flex;
     flex-direction: column;
 
-    & * {
-      line-height: 28px;
-    }
-
-    & p {
-      font-weight: 400;
-    }
-
-    & a {
-      font-weight: 400;
-    }
-
     & li {
       white-space: normal;
-      margin-top: -1.5rem;
+    }
 
-      & li {
+    & ol, & ul {
+      overflow-wrap: anywhere;
+    }
+
+    & h1, & h2, & h3 {
+      & +ol, & +ul {
         margin-top: 0;
+        /* background-color: ${palette.blue.dark2}; */
       }
     }
 
-    & code {
-      white-space: pre-wrap;
-    }
-  `,
-  // End hacky fix
-  markdown_list: css`
-    overflow-wrap: anywhere;
-    margin-bottom: -1.5rem;
-
-    & ul {
-      margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


packages/mongodb-chatbot-ui/src/Message.tsx
```diff
@@ -74,42 +74,23 @@ const styles = {
...
-  markdown_list: css`
-    overflow-wrap: anywhere;
-    margin-bottom: -1.5rem;
 
-    & ul {
-      margin-bottom: 0;
+    & h1, & h2, & h3 {
+      & +ol, & +ul {
+        margin-top: 0;
+        /* background-color: ${palette.blue.dark2}; */
```
mongodben, line 88: will this support both dark and light mode?

mongodben:
i think the spacing between list items now looks strangely big. what was the reason for this change?

packages/mongodb-chatbot-ui/src/Message.tsx
```diff
@@ -153,14 +134,14 @@ const markdownProps = {
     },
     ol: ({ children, ordered, ...props }) => {
       return (
-        <Body as="ol" className={styles.markdown_list} {...props}>
+        <Body as="ol" {...props}>
```
mongodben, line 137: should `styles.markdown_list` be deleted accordingly?

mongodben (changes requested):
couple questions but generally looks good

packages/mongodb-chatbot-ui/src/Message.tsx
nlarew: Meant to delete this line! I used it to highlight the targeted elements in dev.

packages/mongodb-chatbot-ui/src/Message.tsx
nlarew: Yeah it already is. These styles got merged into `markdown_container`.

nlarew:
> i think the spacing between list items now looks strangely big. what was the reason for this change?

@mongodben This is the base styles from LG Chat. Still not sure why our list items spread out this much but I think it's (a) not terrible and (b) less complicated to use the out of the box styles rather than hacking around every edge case.

mongodben (approved):

PR #531, opened 2024-10-04
(EAI-566): Support creating new conversation when add message when conversationId = "null"

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-566

## Changes

- Add logic to create new conversation from `addMessageToConversation` endpoint when conversationId = "null"
- Document functionality
- Update `mongodb-chatbot-server` test suite to use in-memory MongoDB. this reduces risk of us making a collection in dev/test clusters and not cleaning it up. could also affect speed of test suite in CI (hopefully for the better)

## Notes

- This functionality is added to our implementation by default as its the new default behavior


packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -70,6 +63,26 @@ export interface AddMessageToConversationRouteParams {
...
   addMessageToConversationCustomData?: AddCustomDataFunc;
+  /**
+    If present, the route will create a new conversation
+    when given the `conversationIdPathParam` in the URL.
+   */
+  createConversation?: {
+    /**
+      Create a new conversation when the `conversationId` is the string "null".
+     */
+    createOnNullConversationId: boolean;
```
nlarew, line 74: Seems like there's a mismatch here between the type & the description comment. i.e. `conversationIdPathParam` vs `createOnNullConversationId`.

Based on the type, it seems like we're not implementing the more customizable route and just using `/conversations/null/messages`. If that's the case, I think we should omit `createOnNullConversationId` entirely and let the boolean check be directly on whether or not `createConversation` is defined.

When would I want to define this config and then _not_ have it enabled?

packages/mongodb-chatbot-server/src/test/testConfig.ts
```diff
@@ -28,7 +30,26 @@ import {
   makeFilterNPreviousMessages,
 } from "../processors";
 import { makeDefaultReferenceLinks } from "../processors/makeDefaultReferenceLinks";
+import { MongoMemoryServer } from "mongodb-memory-server";
```
nlarew, line 33: How does this interact with some of our Atlas-only functionality, i.e. vector search?

Is the idea that we use the dev db `pages`/`embedded_content` collections for retrieval but store the `conversations` collection in memory?

nlarew (changes requested):

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -249,10 +274,37 @@ async function addMessagesToDatabase({
...
+    });
+    return await conversations.create({
+      initialMessages: createConversation.systemMessage
+        ? [createConversation.systemMessage]
+        : undefined,
+      customData: createConversation.addCustomData
+        ? await createConversation.addCustomData(req, res)
+        : undefined,
+    });
+  }
```
nlarew, line 307: (Comment is for the line below this - not included in PR)

I think we should refactor away from `toObjectId` (we only use it one time, right here) and instead inline the `ObjectId` cast. We should also use a less technical, more descriptive error message in the response.

e.g. `Invalid conversationId: ${id}`

packages/mongodb-chatbot-server/src/routes/conversations/conversationsRouter.test.ts
```diff
@@ -200,6 +200,9 @@ describe("Conversations Router", () => {
     await createConversationReq({ app, origin });
     expect(called).toBe(true);
   });
+  it("should create a new conversation with 'null' value for addMessageToConversation if configured", async () => {
+    // TODO: implement
```
nlarew, line 204: Do we need to implement this? Looks like it may already be covered in the `addMessageToConversation` test.

packages/mongodb-chatbot-server/src/test/testConfig.ts
mongodben: well for the `mongodb-chatbot-server` package, we only really need DB interaction for `conversations`. The core `embedded_content`/vector search logic is in `mongodb-rag-core`. And the `mongodb-chatbot-server` module doesn't directly touch `pages`. 

this PR removes the need for mongodb for `conversations`. We could also in future PR remove need for atlas for `embedded_content` by creating a mock embedded content store or doing something fully in memory like FAISS.

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
mongodben: the reason that this field exists, which is required if you specify `createConversation` is b/c there are other optional fields. 

if there weren't this `createOnNullConversationId` field, then the way that you'd configure create conversation on the endpoint w/o the optional fields would be:

```ts
{
  // ...other config
  createConversation: {/*empty obj*/}
}
```

packages/mongodb-chatbot-server/src/test/testConfig.ts
mongodben: made a ticket for mocking embeddedcontentstore in the mongodb-chatbot-server instead of using atlas vector search. also included removing azure openai from the tests there. azure openai causes tests to be quite long running - https://jira.mongodb.org/browse/EAI-572

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
mongodben: implemented change to inline

nlarew (approved):

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
nlarew: IMO that empty object _is_ a legit way of configuring it (i.e. presence of the field, regardless of the value)

One other small option to consider is allowing `createConversation` to be `boolean | CreateConversationConfig` and then you could do `createConversation: true` instead of the empty object.

mongodben: i'm going to keep this as is, but i understand all your points. i don't think that any of these solutions are optimal, but also think it doesn't really matter and don't want to get lost in the bikeshed here instead of delivering customer value™️

PR #536, opened 2024-10-10
(EAI-575) [UI] Message Metadata Merging

nlarew:
Jira: (EAI-575) [UI] Message Metadata Merging

## Changes

- Adds `updateMessageMetadata` method & uses it in place of `setMessageMetadata`

### DIFF
--- a/packages/mongodb-chatbot-ui/src/useConversation.tsx
+++ b/packages/mongodb-chatbot-ui/src/useConversation.tsx
@@ -76,6 +76,10 @@ type ConversationActor = {
   setMessageMetadata: (
     args: ConversationActorArgs<"setMessageMetadata">
   ) => Promise<void>;
+  updateMessageMetadata: (args: {
+    messageId: string;
+    metadata: AssistantMessageMetadata;
+  }) => Promise<void>;
   deleteMessage: (messageId: string) => Promise<void>;
   rateMessage: (messageId: string, rating: boolean) => Promise<void>;
   commentMessage: (messageId: string, comment: string) => Promise<void>;
@@ -519,7 +523,7 @@ export function useConversation(params: UseConversationParams) {
             references.push(...data);
           },
           onMetadata: async (metadata) => {
-            setMessageMetadata({
+            updateMessageMetadata({
               messageId: STREAMING_MESSAGE_ID,
               metadata,
             });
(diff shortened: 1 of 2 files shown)


packages/mongodb-chatbot-ui/src/useConversation.tsx
```diff
@@ -591,6 +595,22 @@ export function useConversation(params: UseConversationParams) {
...
+    async ({ messageId, metadata }) => {
+      if (!state.conversationId) {
+        console.error(`Cannot updateMessageMetadata without a conversationId`);
+        return;
+      }
+      dispatch({
+        type: "setMessageMetadata",
+        messageId,
+        metadata: {
+          ...state.messages.find((m) => m.id === messageId)?.metadata,
```
mongodben, line 608:
```suggestion
          ...(state.messages.find((m) => m.id === messageId)?.metadata ?? {}),
```

i think you need this so spread operator doesn't break on undefined

packages/mongodb-chatbot-ui/src/useConversation.tsx
```diff
@@ -591,6 +595,22 @@ export function useConversation(params: UseConversationParams) {
...
+          ...metadata,
```
mongodben, line 609: cna metadata be undefined? if so, i think prefer nullish spreading

```suggestion
          ...(metadata ?? {}),
```

mongodben (changes requested):
see notes. also build failing.

also, i think this file really ought to be tested given the amount of core app logic in it.

packages/mongodb-chatbot-ui/src/useConversation.tsx
nlarew: Spread operator supports undefined

`{ ...undefined } => {}`

packages/mongodb-chatbot-ui/src/useConversation.tsx
nlarew: We validate that the event is well-formed & the field exists in the `addMessageStreaming` implementation so this won't be undefined. Also spread operator supports undefined.

nlarew:
> this file really ought to be tested

Fair enough, needs some setup / probably investigation into config stuff

https://jira.mongodb.org/browse/EAI-576

packages/mongodb-chatbot-ui/src/useConversation.tsx
mongodben: noted! i must have been hallucinating...

PR #537, opened 2024-10-16
(EAI-539): Run text-to-Node.js driver benchmark

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-539

## Changes

- support non-radiant models that follow OpenAI spec
- change fuzzymatcher to return `0` instead of `null` if execution fails. easier for reporting
- improvements to handling executing driver code
- fuzzy match improvements
- Logic to pass RUN_ID when executing script
- additional logic to handle different model rate limits, sleep between executions, concurrency, etc.
- Script to materialize top-level data in CSV for report

## Notes

-


packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
```diff
@@ -1,98 +1,139 @@
...
-            promptType,
-            generateCollectionSchemas,
-          });
+    const modelExperiments = models
+
+      .filter((m) => m.authorized === true)
+      // // NOTE: ignoring Google models for now b/c of issues with Radiant
+      .filter((m) => m.developer !== "Google")
+      // TODO: sequentially work thru the models
+      .filter((m) => m.label.includes("llama"))
```
nlarew, line 71: Why do we only run if there is a `llama` label?

Also, is the TODO done or is there an EAI we can add to the comment?

nlarew (approved):
LGTM - functionally I think this is good to go / the benchmark results are sound

Left a few suggestions on code style/API + 1 question on llama models

packages/benchmarks/src/textToDriver/nodeJsDriverBenchmark.ts
mongodben: ah this line can be removed. this was just for me to do while iterating thru

mongodben:
```suggestion
```

packages/benchmarks/src/makeOpenAiClientFactory.ts
```diff
@@ -0,0 +1,101 @@
...
+  azure?: BaseModelProviderConfig & {
+    apiVersion: string;
+  };
+  braintrust?: BaseModelProviderConfig;
+  radiant?: BaseModelProviderConfig & {
+    authCookie: string;
+  };
+}
+
+export function makeOpenAiClientFactory({
```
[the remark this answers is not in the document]
mongodben: i think this factory func will be relevant for future benchmarks that use various models. i prefer this pattern as is b/c it allows for more dynamic instantiation

PR #546, opened 2024-10-30
(EAI-587): Clean up shared imports in mongodb-rag-core + migrate `@azure/openai` to `openai`

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-587

## Changes

- Switch codebase to use `openai` module instead of `@azure/openai`
- Rename modules exported from `mongodb-rag-core` to better folow namespace convention of `CamelCase` (w/ first letter uppercase)
  - Due to small differences in the SDKs' APIs, there were some small associated code refactors.

## Notes

-


packages/mongodb-rag-core/src/index.ts
```diff
@@ -30,5 +29,5 @@ export * from "./assertEnvVars";
 
 // Everyone share the same versions of these packages
 export * from "mongodb";
-export * from "@azure/openai";
+export * as OpenAI from "openai";
 export * from "./langchain";
```
mongodben, line 33: @nlarew in future PR, i wanna do the hared langchain and mongodb exports w/ the `* as Name` convention for continuing the better namespacing of modules in this package,

ie:

```ts
export * as MongoDB from "mongodb";
export * as Langchain from "langchain";
```

not doing in this PR b/c it'd dramatically increase the # of files touched.

relatedly, i think it'd probably be *more correct* to change these external dependencies that we re-export as `peerDependencies` in the package.json, but also since our monorepo is the main consumer of the `mongodb-rag-core` package, it probably would add more overhead and annoyance to us than value

nlarew: I'm directionally ok with that `export * as Namespace` approach, though I want to make sure that it wouldn't make it harder to do tree shaking or otherwise bloat bundles.

If it is an issue, we might consider an alternative of two package entry points. i.e.

-  In this file, we `export * from "./langchain"` so that there's no namespace wrapper object
- We need to create that file we export from, i.e. `./langchain/index.ts`. It does the `export * from "openai";`
- In `package.json` we add a `exports` for `./langchain`.

So basically I could do either of these as a consumer:

```ts
import { LangChainGithubRepoLoader } from "mongodb-rag-core";
import { LangChainGithubRepoLoader } from "mongodb-rag-core/langchain";
```

Maybe too complex?

nlarew: Interesting thought on the peer dependencies, I'd have to think through it a bit or see it in action

mongodben: i'm notionally in agreement w/ the idea of `mongodb-rag-core/<namespace>`, however in practice i've repeatedly run into issues w/ any packages that we compose like this.

mongodben: that's why i didn't use that pattern. but let me use this as an opportunity to at least properly understand why this doesn't work, even if we cant switch to it..tho maybe we'll be able to switch..let's see.

packages/mongodb-rag-core/src/dataSources/MdOnGithubDataSource.test.ts
```diff
@@ -35,13 +36,13 @@ const mongodbCorpConfig: MakeMdOnGithubDataSourceParams = {
...
 
-const ingestTestDataConfig: MakeMdOnGithubDataSourceParams = {
+const testDataConfig: MakeMdOnGithubDataSourceParams = {
   ...baseChatbotRepoConfig,
   name: "ingest_testData",
   metadata: {
     productName: "Ingest Test Data",
   },
-  filter: (path) => path.includes("ingest/testData"),
+  filter: (path) => path.includes(Path.join("mongodb-rag-core", "testData")),
```
nlarew, line 45: Do we need the `Path` import here? Can't you just do `"mongodb-rag-core/testData"` since this is joining two static strings?

packages/chatbot-server-mongodb-public/src/processors/extractMongoDbMetadataFromUserMessage.ts
```diff
@@ -41,7 +40,7 @@ const systemPrompt = `You are an expert data labeler employed by MongoDB.
 You must label metadata about the user query based on its context in the conversation.
 Your pay is determined by the accuracy of your labels as judged against other expert labelers, so do excellent work to maximize your earnings to support your family.`;
 
-const fewShotExamples: ChatCompletionMessageParam[] = [
+const fewShotExamples: OpenAI.default.Chat.ChatCompletionMessageParam[] = [
```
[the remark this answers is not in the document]
mongodben: yeah, also the `openai` package does some funky stuff w/ exports

packages/mongodb-rag-core/src/dataSources/MdOnGithubDataSource.test.ts
mongodben: not if you're on windows!

packages/mongodb-chatbot-server/src/routes/generateResponse.ts
```diff
@@ -458,16 +460,21 @@ export async function streamGenerateResponseMessage({
     });
   }
 
-  return { messages: newMessages };
+  return { messages: newMessages.map(convertMessageFromLlmToDb) };
```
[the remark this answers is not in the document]
mongodben: b/c in our implementation it's still type `functionCall`, whereas the new openai lib is `function_call`

packages/chatbot-server-mongodb-public/src/test/mockOpenAi.ts
```diff
@@ -1,25 +1,33 @@
...
-                message: {
-                  tool_calls: [
-                    {
-                      function: {
-                        arguments: JSON.stringify(funcRes),
-                      },
+export const makeMockOpenAIToolCall = (funcRes: Record<string, unknown>) => {
+  // This module is also required for the mock to work
+  // in the tests
+  const { updateFrontMatter } = jest.requireActual("mongodb-chatbot-server");
```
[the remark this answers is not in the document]
mongodben: yea b/c that's needed in one of the methods. don't want it overwritten

packages/mongodb-rag-core/src/index.ts
mongodben: going to merge this PR, and play w/ this in separate PR. would require touching a lot of files b/c so many imports would be changed

packages/mongodb-artifact-generator/src/chat/makeClassifier.ts
```diff
@@ -150,28 +146,29 @@ export function makeClassifier({
...
+    });
     const response = result.choices[0].message;
     if (response === undefined) {
       throw new Error("No response from OpenAI");
     }
-    if (response.functionCall === undefined) {
+    if (
+      response.function_call === undefined ||
+      response.function_call === null
+    ) {
```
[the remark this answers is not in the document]
mongodben: keeping explicit in case of `""`

packages/chatbot-server-mongodb-public/src/test/mockOpenAi.ts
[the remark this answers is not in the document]
nlarew: I guess the meat of the question is - when would we need to add other methods here?

PR #547, opened 2024-10-31
(EAI-466): move common modules from `mongodb-artifact-generator` to `mongodb-rag-core`

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-466

## Changes

- move `makeClassifier`
-

## Notes

- note: in the jira issue, there was a note about including some general 'translate' or 'summarize' LLM functions, but i didn't see any generalizable implementations to include.

### DIFF
--- a/packages/mongodb-rag-core/src/makeClassifier.ts
+++ b/packages/mongodb-rag-core/src/makeClassifier.ts
@@ -1,8 +1,3 @@
-import "dotenv/config";
-import {
-  assertEnvVars,
-  CORE_OPENAI_CHAT_COMPLETION_ENV_VARS,
-} from "mongodb-rag-core";
 import {
   ChatRequestMessage,
   FunctionDefinition,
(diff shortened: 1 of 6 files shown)


packages/mongodb-rag-core/src/makeClassifier.ts
```diff
@@ -173,19 +167,6 @@ export function makeClassifier({
...
-        <SystemMessage>
-          ${messages[0].content}
-        </SystemMessage>
-        <Classification>
-          ${JSON.stringify(classification)}
-        </Classification>
-      `
-    );
-
     return classification;
```
mongodben, line 170: @nlarew i started working on moving the `makeClassifier()` to `mongodb-rag-core`, but the classifier is tightly coupled w/ the RunLogger, that i'm not sure how to do this elegantly.

curious your thoughts how to proceed w/ this

nlarew: Yeah good point this probably should not be coupled and instead by more functional. I can do that separately or you can do it here.

The gist would be removing the logger call from the `makeClassifier()` implementation and instead putting it in the Jira Prompt Response command’s logic after we call the classifier.

mongodben: > The gist would be removing the logger call from the makeClassifier() implementation and instead putting it in the Jira Prompt Response command’s logic after we call the classifier.

that seems reasonable to me. i'm going to close this PR and remove the ticket from the core consolidation epic since it'd require some more refactoring

PR #548, opened 2024-10-31
Support `/exports` from `mongodb-rag-core`

mongodben:
Jira: n/a

## Changes

- Switch repo to support `/exports`. To make this work, I had to add the following to the root `tsconfig.json`:

```jsonc
{
    // ...
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    // ...
}
```
- Update imports to use new `/export` `mongodb-rag-core` package
- Remove `chatbot-eval-mongodb-public` (it's unused)
  - Note: was unable to use `mongodb-chatbot-evaluation` b/c it has some modules used in benchmarks. we should remove that and move the modules over in a separate PR.

## Notes

-


packages/benchmarks/src/makeRadiantChatLlm.ts
```diff
@@ -15,14 +16,14 @@ export const makeRadiantChatLlm = async ({
   deployment: string;
   mongoDbAuthCookie?: string;
   lmmConfigOptions: Pick<
-    OpenAI.default.Chat.ChatCompletionCreateParams,
+    OpenAI.Chat.ChatCompletionCreateParams,
```
nlarew, line 19: I don't dislike the namespaced types here (very explicit what it is!). Is it _possible_ to import just `ChatCompletionCreateParams` without the whole `OpenAI` namespace?

packages/mongodb-chatbot-server/src/test/testConfig.ts
```diff
@@ -47,7 +48,7 @@ const allowedOrigins = process.env.ALLOWED_ORIGINS?.split(",") || [];
   Boost results from the MongoDB manual so that 'k' results from the manual
   appear first if they exist and have a min score of 'minScore'.
  */
-export const boostManual = makeBoostOnAtlasSearchFilter({
+export const boostManual: SearchBooster = makeBoostOnAtlasSearchFilter({
```
nlarew, line 51: Why the explicit type annotation? Shouldn't `makeBoostOnAtlasSearchFilter` already resolve to a `SearchBooster` based on its signature?

packages/scripts/src/TypeChatJsonTranslateFunc.ts
```diff
@@ -48,13 +49,12 @@ export function makeTypeChatJsonTranslateFunc<SchemaType extends object>({
     apiKey,
     `${baseUrl}openai/deployments/${deployment}/chat/completions?api-version=${version}`
   );
-
-  // LLM function
-  const translator = createJsonTranslator<SchemaType>(
-    model,
+  const validator = createTypeScriptJsonValidator<SchemaType>(
```
nlarew, line 52: New TypeChat update? I don't see a version bump or anything.

tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "target": "es2020",
     "lib": ["ES2021", "DOM"],
-    "module": "CommonJS",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
nlarew, line 6: There are some considerable implications in changing the module format. I think this is something we should do, but we'll want to monitor for impact + maybe do a major version bump

nlarew (approved):
LGTM % a few questions throughout

packages/chatbot-server-mongodb-public/src/processors/makeStepBackUserQuery.test.ts
```diff
@@ -1,13 +1,13 @@
 import { makeMockOpenAIToolCall } from "../test/mockOpenAi";
 import { makeStepBackUserQuery } from "./makeStepBackUserQuery";
-import { OpenAI } from "mongodb-chatbot-server";
-jest.mock("mongodb-chatbot-server", () =>
+import { OpenAI } from "mongodb-rag-core/openai";
+jest.mock("mongodb-rag-core/openai", () =>
```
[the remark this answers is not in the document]
mongodben: needed b/c now mongodb-chatbot-server is no longer exporting the /openai module

packages/scripts/src/TypeChatJsonTranslateFunc.ts
mongodben: yeah

packages/chatbot-server-mongodb-public/src/test/mockOpenAi.ts
```diff
@@ -1,33 +1,29 @@
 export const makeMockOpenAIToolCall = (funcRes: Record<string, unknown>) => {
   // This module is also required for the mock to work
   // in the tests
-  const { updateFrontMatter } = jest.requireActual("mongodb-chatbot-server");
```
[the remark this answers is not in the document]
mongodben: def delete comment too. will implement

packages/scripts/src/TypeChatJsonTranslateFunc.ts
nlarew: I don't see a version bump in the diff - do we need to update `scripts/package.json`?

mongodben: maybe? i'll set the version in package.json to whatever is most recent

packages/benchmarks/src/makeRadiantChatLlm.ts
mongodben: i dont think so. dont think it's a real issue tho b/c the benchmarks package is only being built in this repo, which will already install all of `mongodb-rag-core`, which is dependent on the `openai` package

packages/mongodb-chatbot-server/src/test/testConfig.ts
mongodben: removing

packages/scripts/src/TypeChatJsonTranslateFunc.ts
mongodben: version in package.json is the latest. i'm confused what's going on, but will keep as is since it works and i dont think this usage is important enough for us to worry too much about

PR #567, opened 2024-12-02
(EAI-619) [UI] Make invalid stream event types a no-op

nlarew:
Jira: (EAI-619) [UI] Make invalid stream event types a no-op

## Changes

- We're now more relaxed if the server streams an SSE event that doesn't match our expectations. We used to throw, now we just log an error (in non-prod envs)

### DIFF
--- a/packages/mongodb-chatbot-ui/src/services/conversations.ts
+++ b/packages/mongodb-chatbot-ui/src/services/conversations.ts
@@ -299,7 +299,6 @@ export class ConversationService {
 
       onmessage(ev) {
         const event = JSON.parse(ev.data);
-        console.log("onmessage", event);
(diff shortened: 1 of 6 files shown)


packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -280,8 +299,28 @@ export class ConversationService {
 
       onmessage(ev) {
         const event = JSON.parse(ev.data);
+        console.log("onmessage", event);
```
mongodben, line 302:
```suggestion
```

should be removed?

mongodben (approved):
lgtm, see small suggestion

PR #570, opened 2024-12-04
Support AWS bedrock models in `benchmarks` package

mongodben:
Jira: n/a

## Changes

- Support AWS bedrock models. Includes the following:
  - Previously through Radiant, now directly through Bedrock:
    - Anthropic: Claude 3 Sonnet|Haiku, Claude 3.5 Sonnet
    - Llama: 3 70b
  - New models:
    - Anthropic: Claude 3.5 Sonnet v2, Claude 3.5 Haiku
    - Llama: 3.1 70b, 3.2 90b
    - Amazon Nova (released yesterday!): micro, lite, pro
 
## Notes

- Note that support limited to chat completions without streaming


packages/benchmarks/src/makeOpenAiClientFactory.ts
```diff
@@ -65,6 +75,70 @@ export function makeOpenAiClientFactory({
...
+            content: responseText,
+          },
+        },
+      ],
+    } as unknown as Awaited<
+      ReturnType<typeof openAiClient.chat.completions.create>
+    >;
+  };
+
+  // Override the `.create()` method with minimal type casting
```
mongodben, line 135:
```suggestion
```

packages/benchmarks/.env.example
```diff
@@ -18,3 +18,7 @@ OPENAI_API_KEY="<api_key>"
 MONGODB_AUTH_COOKIE="<Auth cookie from CorpSecure-protected site>"
 GCP_OPENAI_ENDPOINT="https://{LOCATION}-aiplatform.googleapis.com/v1beta1/projects/{PROJECT_ID}/locations/{LOCATION}/endpoints/openapi"
 GCP_API_KEY="SOME_API_KEY"
+AWS_REGION="us-east-1"
+AWS_ACCESS_KEY_ID="..."
+AWS_SECRET_ACCESS_KEY="..."
+AWS_SESSION_TOKEN="note: must refresh regularly"
```
nlarew, line 24: Is the token something we could fetch programmatically through e.g. STS?

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/sts/command/GetSessionTokenCommand/

packages/benchmarks/src/models.ts
```diff
@@ -110,66 +112,98 @@ export const models: ModelConfig[] = [
...
+    authorized: true,
+  },
+  {
+    label: "nova-pro-v1:0",
+    deployment: "amazon.nova-pro-v1:0",
+    developer: "Amazon",
+    provider: "aws_bedrock",
+    authorized: true,
   },
+  // Note: not available in Bedrock (yet?)
```
nlarew, line 206: Do we need a follow up EAI ticket?

nlarew (approved):

packages/benchmarks/src/models.ts
mongodben: no, i dont think so. like bedrock doesn't support it currently. 

when bedrock adds more models, we can add them. 

i dont think it's a big deal that we're missing this one specifically

packages/benchmarks/.env.example
mongodben: theres def some other ways we could get this token, which we'd wanna use for longer runnign tasks.

i think the current default token expiration time is like 1 hour

PR #578, opened 2024-12-12
Add braintrust to mongodb-rag-core

mongodben:
Jira: n/a

## Changes

- Add `braintrust` as dependency to mongodb-rag-core, and use throughout project. doing this b/c we will need to integrate it with `mongodb-chatbot-server` for observability. this'll make 3 packages dependent on the library, so easier to maintain if we keep dependency centralized to core. 
-

## Notes

-


packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -30,7 +30,6 @@ import { GenerateUserPromptFunc } from "../../processors/GenerateUserPromptFunc"
 import { FilterPreviousMessages } from "../../processors/FilterPreviousMessages";
 import { filterOnlySystemPrompt } from "../../processors/filterOnlySystemPrompt";
 import { generateResponse } from "../generateResponse";
```
mongodben, line 32:
```suggestion
import { generateResponse } from "../generateResponse";

```

nlarew (approved):

PR #579, opened 2024-12-12
(EAI-622): Add braintrust tracing to server

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-622

## Changes

-
-

## Notes

-


packages/chatbot-server-mongodb-public/src/tracing.ts
```diff
@@ -0,0 +1,84 @@
...
+) => AppConfig["conversationsRouterConfig"]["addMessageToConversationUpdateTrace"] = (
+  k
+) =>
+  async function ({ traceId, addedMessages, logger }) {
+    const tracingData = extractTracingData(addedMessages);
+    logger.updateSpan({
+      id: traceId,
+      tags: tracingData.tags,
+      scores: {
+        RejectedQuery: tracingData.rejectQuery === true ? 1 : null,
```
nlarew, line 16: Curious - is there a meaningful difference between a score of `null` and a score of `0`?

packages/mongodb-rag-core/src/conversations/MongoDbConversations.ts
```diff
@@ -164,9 +164,9 @@ export function makeMongoDbConversationsService(
 
 export function createMessage(messageParams: AddSomeMessageParams) {
   const message = {
-    id: new ObjectId(),
-    createdAt: new Date(),
     ...messageParams,
+    createdAt: new Date(),
+    id: messageParams.id ?? new ObjectId(),
```
nlarew, line 169: Why this switch? What if we want to override the current date for `createdAt`, e.g. in a test (not actually sure if/when we do this)

nlarew (approved):

packages/chatbot-server-mongodb-public/src/tracing.ts
mongodben: yep, `null` means 'no score registered', whereas 0 is the lowest possible score

packages/mongodb-rag-core/src/conversations/MongoDbConversations.ts
mongodben: it was for capturing the messageParams.id, which is needed for the tracing

PR #580, opened 2024-12-16
check for change in chunkAlgoHash

yakubova92:

## Changes

-
-

## Notes

-

### DIFF
--- a/packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
+++ b/packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
@@ -18,6 +18,21 @@ export interface EmbedConcurrencyOptions {
   createChunks?: number;
 }
 
+const chunkAlgoHashes = new Map<string, string>();
+
+const getHashForFunc = (f: ChunkFunc, o?: Partial<ChunkOptions>): string => {
+  const data = JSON.stringify(o ?? {}) + f.toString();
+  const existingHash = chunkAlgoHashes.get(data);
+  if (existingHash) {
+    return existingHash;
+  }
+  const hash = createHash("sha256");
+  hash.update(data);
+  const digest = hash.digest("hex");
+  chunkAlgoHashes.set(data, digest);
+  return digest;
+};
+
 /**
   (Re-)embeddedContent the pages in the page store that have changed since the given date
   and stores the embeddedContent in the embeddedContent store.
@@ -48,6 +63,23 @@ export const updateEmbeddedContent = async ({
       sourceNames ? ` in sources: ${sourceNames.join(", ")}` : ""
     }`
   );
+  // find changed chunkingHashes 
+  const allPages = await pageStore.loadPages({sources: sourceNames});
+  for (const page of allPages) {
+    const existingContent = await embeddedContentStore.loadEmbeddedContent({
+      page,
+    });
+    const chunkAlgoHash = getHashForFunc(chunkPage, chunkOptions);
+    if (existingContent[0].chunkAlgoHash !== chunkAlgoHash) {
+      logger.info(
+        `Chunking algorithm has changed for ${page.sourceName}: ${page.url}. Deleting existing embedded content to force an update.`
+      );
+      await embeddedContentStore.deleteEmbeddedContent({
+        page,
+      });
+    }
+  }
+
   await PromisePool.withConcurrency(concurrencyOptions?.processPages ?? 1)
     .for(changedPages)
     .process(async (page, index, pool) => {
@@ -71,21 +103,6 @@ export const updateEmbeddedContent = async ({
           });
       }
     });
-};
-
-const chunkAlgoHashes = new Map<string, string>();
-
-const getHashForFunc = (f: ChunkFunc, o?: Partial<ChunkOptions>): string => {
-  const data = JSON.stringify(o ?? {}) + f.toString();
-  const existingHash = chunkAlgoHashes.get(data);
-  if (existingHash) {
-    return existingHash;
-  }
-  const hash = createHash("sha256");
-  hash.update(data);
-  const digest = hash.digest("hex");
-  chunkAlgoHashes.set(data, digest);
-  return digest;
 };
 
 export const updateEmbeddedContentForPage = async ({


packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
```diff
@@ -18,6 +18,21 @@ export interface EmbedConcurrencyOptions {
   createChunks?: number;
 }
 
+const chunkAlgoHashes = new Map<string, string>();
```
mongodben, line 21: this is to cache existing hashes so you dont have to recompute? 

also preferable to scope this within a function `updateEmbeddedContent()`. better not to instantiate things at the global scope if you can avoid. this way it's not instantiated til the func is called and its cleaned up after the func's run is over then pass the instantiation to the getHashForFunc() func. pseduocode:

function getHashForFunc(f, o, cache) {
// ...
}

function updateEmbeddedContent(/* args*/) {
  const chunkAlgoHashes = new Map<string, string>();
  // do stuff
  const chunkAlgoHash = getHashForFunc(chunkPage, chunkOptions, chunkAlgoHashes);
}

mongodben (changes requested):
see comment about caching.

and beyond that, sorry if i'm dense, but what changes are you making?

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
yakubova92: I see your point and can make that change. This was existing code that I simply moved up (kept same scope) because I was calling it before it was defined.

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
```diff
@@ -48,6 +63,18 @@ export const updateEmbeddedContent = async ({
...
+  const allPages = await pageStore.loadPages({sources: sourceNames});
+  const pagesWithChangedChunking = allPages.filter(async (page) => {
+    const existingContent = await embeddedContentStore.loadEmbeddedContent({
+      page,
+    });
+    const chunkAlgoHash = getHashForFunc(chunkPage, chunkOptions);
+    return existingContent[0].chunkAlgoHash !== chunkAlgoHash
+  });
+
+  changedPages.push(...pagesWithChangedChunking);
```
mongodben, line 76: this is very inefficient and uses a lot of memory b/c it loads all pages, loops thru them one by one. pls reconsider a different implementation

mongodben: i think you can use the DB more here to make this logic happen at the DB level. 

you could: 


add a method to the embedded content collection that finds the data sources for a given query, eg:

```ts

// typing 
type EmbeddedContentStore {
  getDataSources(matchQuery: Document): Promise<string[]>
}

// implementation w mongodb
{
  async getDataSources(matchQuery) {
    const {uniqueSources } = await embeddedContentStore.aggregate([

  /* example match query:
{
    "$match": {
      "chunkAlgoHash": "49d78a1d6b12ee6f433a2156060ed5ebfdefb8a90301f1c2fb04e4524944c5eb",
      "sourceName": {
        "$in": [
          "snooty-cloud-docs",
          "snooty-docs"
        ]
      }
    }
  },

  */
  { $match: matchQuery },
  {
    "$group": {
      "_id": null,
      "uniqueSources": {
        "$addToSet": "$sourceName"
      }
    }
  },
  {
    "$project": {
      "_id": 0,
      "uniqueSources": 1
    }
  }
]).toArray();
return uniqueSources;
  }
}

```

you can then merge this list of sources that require update before the initial call to pageStore.loadPages() on line 57:


pseudocode:

```ts
const chunkAlgoHashes = new Map<string, string>();
const chunkAlgoHash = getHashForFunc(chunkPage, chunkOptions, chunkAlgoHashes);

const sourcesWithNewChunkAlgoHash = await embeddedContentStore.getDataSources({
      chunkAlgoHash,
      // just run on specific source names if specified, run on all if not
      ...(sourceNames ? {"sourceName": {
        "$in": sourceNames
      }}: undefined)
    })

const allSourcesToUpdate = new Array(new Set([...sourcesWithNewChunkAlgoHash, sourceNames]])); //smthn like this..not 100% sure on syntax
const changedPages = await pageStore.loadPages({
    updated: since,
    sources: allSourcesToUpdate,
  });

// ... rest of code unchanged

```

mongodben: you may also want to add an index to 'chunkAlgoHash' on the embeddedContent collection to improve query performance (not that it particularly matters here since this is just run once per CLI inovcation)

mongodben: note that this implementation wouldn't be sufficient if we allowed for updating chunks based on criteria besides `dataSource`...but that's not how the CLI currently works, so this implementation should get the job done

yakubova92: refactored to get pages based on embedded content with an aggregation pipeline that uses a $lookup

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -95,6 +95,68 @@ export function makeMongoDbEmbeddedContentStore({
...
+          $match: {
+            ...(dataSources ? { sourceName: { $in: dataSources } } : undefined),
+            chunkAlgoHash: inverseChunkAlgoHash
+              ? { $ne: chunkAlgoHash }
+              : chunkAlgoHash,
+          },
+        },
+        {
+          $lookup: {
+            from: "pages",
```
mongodben, line 114: one potential issue w/ this is that the `PageStore` collection might not be named `Pages`. also by not passing the `PageStore`, you're coupling this method to MongoDB and the 2 collections existing in the same, which shouldn't necessarily be assumed.

for this reason, an implementation more like the one discussed in my comment here could be preferable https://github.com/mongodb/chatbot/pull/580#discussion_r1887467877

mongodben (changes requested):
see comment about the $lookup implementation

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -232,5 +232,22 @@ export function makeMongoDbEmbeddedContentStore({
...
+              _id: null,
+              uniqueSources: { $addToSet: "$sourceName" },
+            },
+          },
+          { $project: { _id: 0, uniqueSources: 1 } },
+        ])
+        .toArray();
+      const uniqueSources = result.length > 0 ? result[0].uniqueSources : [];
+      return uniqueSources;
+    },
```
mongodben, line 251: this implementation goes back to what we've discussed before where the typing isn't correct to make the interface non-dependent on the implementation. 

for example in the implementation in this PR, you're using MDB methods like `$ne` (https://github.com/mongodb/chatbot/pull/580/files#diff-1c357ff4fa8be764f60aa65a2d727938d376bd13e480049c8e7f73379a208b25R78-R86). this makes the implementation dependent on mongodb, whereas thats not what we want to do.

please refactor accordingly to make it so that it can take typed arguments. then you can add the mongodb specific stuff to the implementation.

roughly like:

```ts
interface GetSourcesMatchParams {
  sourceNames?: string[];
  chunkAlgoHash: {
    hashValue: string;
    operation: "equals" | "notEquals"
  }
}

function makeMatchQuery ({sourceName, chunkAlgoHash}: GetSourcesMatchQuery) {

// something like this
let operator = '';
if(chunkAlgoHash.operation === 'equals'){
  operator = '$eq'
} 
if(chunkAlgoHash.operation === 'notEquals'){
  operator = '$ne'
} 

return ({
      chunkAlgoHash: { [operator]: chunkAlgoHash.hashValue },
      // run on specific source names if specified, run on all if not
      ...(sourceNames
        ? {
            sourceName: {
              $in: sourceNames,
            },
          }
        : undefined),
    });
} 
```

```suggestion
    async getDataSources(matchParams: GetSourcesMatchParams): Promise<string[]> {
      const result = await embeddedContentCollection
        .aggregate([
          { $match: makeMatchQuery(matchParams) },
          {
            $group: {
              _id: null,
              uniqueSources: { $addToSet: "$sourceName" },
            },
          },
          { $project: { _id: 0, uniqueSources: 1 } },
        ])
        .toArray();
      const uniqueSources = result.length > 0 ? result[0].uniqueSources : [];
      return uniqueSources;
    },
```

mongodben (changes requested):
see comment about typing

packages/mongodb-rag-core/src/contentStore/EmbeddedContent.ts
```diff
@@ -114,4 +122,9 @@ export type EmbeddedContentStore = VectorStore<EmbeddedContent> & {
     Initialize the store.
    */
   init?: () => Promise<void>;
+
+  /**
+   Get the data sources that match the given query.
```
nlarew, line 127:
```suggestion
   Get the names of ingested data sources that match the given query. 
```

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -58,6 +62,27 @@ export type MongoDbEmbeddedContentStore = EmbeddedContentStore &
...
   };
 
+function makeMatchQuery({ sourceNames, chunkAlgoHash }: GetSourcesMatchParams) {
+  let operator = "";
+  if (chunkAlgoHash.operation === "equals") {
+    operator = "$eq";
+  }
+  if (chunkAlgoHash.operation === "notEquals") {
+    operator = "$ne";
+  }
```
nlarew, line 72:
```suggestion
  const operator = chunkAlgoHash.operation === "equals" ? "$eq" : "$ne";
```
Could simplify to a one-liner here since there are only two options

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.test.ts
```diff
@@ -276,3 +295,231 @@ describe("updateEmbeddedContent", () => {
     });
   });
 });
+
+// These tests use "mongodb-memory-server", not mockEmbeddedContentStore
+describe("updateEmbeddedContent", () => {
```
nlarew, line 300: We shouldn't have two `describe` blocks with the same title. Can you either combine into a single block or disambiguate them with more descriptive titles?

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
```diff
@@ -18,9 +18,26 @@ export interface EmbedConcurrencyOptions {
...
+  hash.update(data);
+  const digest = hash.digest("hex");
+  chunkAlgoHashes.set(data, digest);
+  return digest;
+};
+
 /**
-  (Re-)embeddedContent the pages in the page store that have changed since the given date
-  and stores the embeddedContent in the embeddedContent store.
+  (Re-)embeddedContent the pages in the page store that have changed (copy change, or chunking algothrim change)
```
nlarew, line 39:
```suggestion
  (Re-)embeddedContent the pages in the page store that have changed (content and/or chunking algorithm change) 
```

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
```diff
@@ -39,15 +56,47 @@ export const updateEmbeddedContent = async ({
...
+      sourceNames ? ` in sources: ${sourceNames.join(", ")}` : ""
+    }`
+  );
+  // find all pages with embeddings created using chunkingHashes different from the current chunkingHash
+  const chunkAlgoHashes = new Map<string, string>();
+  const chunkAlgoHash = getHashForFunc(
+    chunkAlgoHashes,
+    chunkPage,
+    chunkOptions
+  );
```
nlarew, line 75: I'm not sure I understand the purpose of the `chunkAlgoHashes` caching here.

* We define an empty cache and only use it for this one call right after.
* That call only gets the current `chunkPage` function (that we import into this file)

It seems like the cache will never actually be used? Am I missing something?

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.ts
```diff
@@ -18,9 +18,26 @@ export interface EmbedConcurrencyOptions {
...
+  const digest = hash.digest("hex");
+  chunkAlgoHashes.set(data, digest);
+  return digest;
+};
+
 /**
-  (Re-)embeddedContent the pages in the page store that have changed since the given date
-  and stores the embeddedContent in the embeddedContent store.
+  (Re-)embeddedContent the pages in the page store that have changed (copy change, or chunking algothrim change) 
+  since the given date and stores the embeddedContent in the embeddedContent store.
```
nlarew, line 40: "since the given date" is only true if there was a content change, right? IIUC a change to our chunk function/options will ignore the since date?

Might need a reword to reflect that.

nlarew (changes requested):
A few small suggested tweaks + one question about the caching behavior

packages/mongodb-rag-core/src/contentStore/updateEmbeddedContent.test.ts
yakubova92: I created a more specific describe block for this for now, but this test file should be refactored. These test cases were placed in their own describe block bc the original describe block used mocks for the page and embedding stores and we are moving to using mongodb-memory-server. I created a ticket to capture that work - https://jira.mongodb.org/browse/EAI-935

mongodben (approved):

PR #582, opened 2024-12-20
(EAI-616)split test case into more isolated test cases

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-616

## Changes

- test case "handles embedded content" was removed
- test cases for loading and updating embedded content were added.

- I noticed the test case "has an overridable default collection name" was listed twice, I removed one

## Notes

-


packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.test.ts
```diff
@@ -39,46 +64,9 @@ describe("MongoDbEmbeddedContentStore", () => {
...
-        tags: [],
-      },
-      updated: new Date(),
-      url: "/a/b/c",
-    };
-
-    const pages = [page, anotherPage];
     for (const page of pages) {
-      const embeddedContent = await store.loadEmbeddedContent({ page });
+      embeddedContent = await store.loadEmbeddedContent({ page });
```
nlarew, line 68: (Applies to the following line - can't comment on it directly)

We shouldn't have `expect()` calls outside of `test/it` blocks. This for loop used to be in the `it("handles embedded content", ...` block but that was removed and now it's inside of `beforeEach`.

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.test.ts
```diff
@@ -139,87 +122,43 @@ describe("MongoDbEmbeddedContentStore", () => {
...
+        })
+      ).toStrictEqual([]);
+      expect(
+        await store.loadEmbeddedContent({
+          page: { ...page, url: page.url + "/" },
+        })
+      ).toStrictEqual([]);
+    });
+  });
+  it("updateEmbeddedContent replaces existing embedded content belonging to the page provided with the new embedded content array provided", async () => {
```
nlarew, line 149: We have a describe block for `loadEmbeddedContent` and `deleteEmbeddedContent` that has individual test cases within.

Let's use a similar structure here, i.e. a describe block for `updateEmbeddedContent` with this test inside.

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.test.ts
```diff
@@ -159,67 +111,45 @@ describe("MongoDbEmbeddedContentStore", () => {
...
-            sourceName: page.sourceName,
-            text: page.body,
-            url: page.url,
-          },
-        ]);
-      }
+      const pageEmbeddings = await store.loadEmbeddedContent({ page });
+      const anotherPageEmbeddings = await store?.loadEmbeddedContent({
+        page: anotherPage,
+      });
```
nlarew, line 120: We should either use optional chaining (`?.`) on both of these call or neither

nlarew (commented):
1 nitpicky request + need to resolve merge conflicts, then this should be good to go

nlarew (approved):

PR #583, opened 2025-01-04
(EAI-614) add init action to ingest:pages and ingest:embed options

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-614

## Changes

- init added to ingest:pages
- init added to ingest:embed
- init calls removed from withConfig function

## Notes

- I modified the npm commands for `ingest:pages` and `ingest:embed` to be able to run this locally. I can remove this, but leaving it in for now so that I remember to address it. It might make sense to set up `ingest:<pages|embed>:update` if we need a one liner for the update action, and have `ingest:<pages|embed>` set up to accept actions. I could use some context on when these commands are called before making that decision.

### DIFF
--- a/packages/ingest-mongodb-public/package.json
+++ b/packages/ingest-mongodb-public/package.json
@@ -5,6 +5,7 @@
   "author": "",
   "license": "Apache-2.0",
   "keywords": [],
+  "main": "build/config.js",
   "engines": {
     "node": ">=18",
     "npm": ">=8"
@@ -18,8 +19,8 @@
     "lint:fix": "npm run lint -- --fix && prettier ./src --check --write",
     "ingest:all": "ingest all --config ./build/config.js",
     "ingest:k8s": "../../node_modules/mongodb-rag-ingest/build/main.js all --config ./build/config.js",
-    "ingest:pages": "ingest pages update --config ./build/config.js",
-    "ingest:embed": "ingest embed update --config ./build/config.js",
+    "ingest:pages": "ingest pages --config ./build/config.js",
+    "ingest:embed": "ingest embed --config ./build/config.js",
     "ingest:pages:meta": "ingest pages update --config ./build/meta.config.js",
     "ingest:k8s:coachGtmPages": "../../node_modules/mongodb-rag-ingest/build/main.js pages update --config ./build/coachGtm.config.js",
     "test": "node --experimental-vm-modules ../../node_modules/jest/bin/jest.js --forceExit"
--- a/packages/ingest-mongodb-public/package.json
+++ b/packages/ingest-mongodb-public/package.json
@@ -18,8 +18,8 @@
     "lint:fix": "npm run lint -- --fix && prettier ./src --check --write",
     "ingest:all": "ingest all --config ./build/config.js",
     "ingest:k8s": "../../node_modules/mongodb-rag-ingest/build/main.js all --config ./build/config.js",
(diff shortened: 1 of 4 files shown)


packages/ingest-mongodb-public/package.json
```diff
@@ -18,8 +18,8 @@
     "lint:fix": "npm run lint -- --fix && prettier ./src --check --write",
     "ingest:all": "ingest all --config ./build/config.js",
     "ingest:k8s": "../../node_modules/mongodb-rag-ingest/build/main.js all --config ./build/config.js",
-    "ingest:pages": "ingest pages update --config ./build/config.js",
-    "ingest:embed": "ingest embed update --config ./build/config.js",
+    "ingest:pages": "ingest pages --config ./build/config.js",
+    "ingest:embed": "ingest embed --config ./build/config.js",
```
nlarew, line 22: What happens when you run `ingest <pages|embed>` without the `init|update` subcommands?

I _think_ the right move here is to have each script call two commands - `init` and then `update`, i.e.

```json
"ingest:pages": "ingest pages init --config ./build/config.js && ingest pages update --config ./build/config.js",
```

packages/mongodb-rag-ingest/src/commands/embed.ts
```diff
@@ -60,6 +67,12 @@ const commandModule: CommandModule<
 
 export default commandModule;
 
+export const doInitEmbedCommand = async ({ embeddedContentStore }: ResolvedConfig) => {
+  if (embeddedContentStore) {
+    await embeddedContentStore?.init?.();
+  }
```
nlarew, line 73: Should we throw an error if the embedded content store isn't defined?

packages/mongodb-rag-ingest/src/commands/pages.ts
```diff
@@ -48,6 +55,12 @@ const commandModule: CommandModule<
 
 export default commandModule;
 
+export const doInitPagesCommand = async ({ pageStore }: ResolvedConfig) => {
+  if (pageStore) {
+    await pageStore?.init?.();
+  }
```
nlarew, line 61: Should we throw an error if the page store isn't defined?

packages/mongodb-rag-ingest/src/commands/embed.ts
mongodben: i think we should

packages/mongodb-rag-ingest/src/commands/pages.ts
mongodben: i think we should

packages/ingest-mongodb-public/package.json
mongodben: i'm confused why you'd need to update this to run it locally

mongodben: also agree with nick's comment

yakubova92: Maybe Im not running it correctly? 
If I run `npm run ingest:pages init` or `npm run ingest:pages delete -- --source mongodb-university` from `ingest-mongodb-public` while the package.json has the script  `ingest:pages` defined as "ingest pages update --config ./build/config.js", then the update will run, when I intend to run init or delete. 

I agree that if the intention of the `ingest:pages` script is to run the update, then it should be updated to init & update as @nlarew suggested.

mongodben (approved):

PR #591, opened 2025-01-14
(EAI-648): Add datasets starter code after port from repo

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-648

## Changes

- Port code from private `mongodb-developer/code-example-dataset` repo
- Do some light formatting to make fit w/ our repo structyre

## Notes
 
- This is very much not production code. Just starting point from the previous work. At a future point, we should clean this all up after borrowing the core logic from it.


nlarew (approved):
Generally LGTM though there's quite a bit here. This is more or less a 1:1 port of the other code examples repo, right?

PR #592, opened 2025-01-14
(EAI-611) chatbot server 500 errs

nlarew:
Jira: (EAI-611) chatbot server 500 errs

## Changes

- Replaces the default Express router with [express-promise-router](https://www.npmjs.com/package/express-promise-router), which handles errors from async routes.

This behavior is baked into `express@^5.0` however I was unable to cleanly upgrade due to some type errors. In the interest of moving forward we can stay on Express 4 and use this instead. Longer term we should debug and upgrade to the latest version. https://jira.mongodb.org/browse/EAI-650


packages/mongodb-rag-core/src/DataStreamer.ts
```diff
@@ -135,11 +136,10 @@ export function makeDataStreamer(): DataStreamer {
...
       }
       sse = makeServerSentEventDispatcher<SomeStreamEvent>(res);
-      // If the client closes the connection, stop sending events
       res.on("close", () => {
-        if (this.connected) {
-          this.disconnect();
-        }
+        logger.info(
+          "Client closed the streaming connection before the stream was finished."
+        );
```
mongodben, line 142: does it disconnect automatically now?

packages/mongodb-chatbot-server/src/utils.ts
```diff
@@ -53,7 +53,9 @@ export const sendErrorResponse = ({
     message: stripIndent`Responding with ${httpStatus} status and error message: ${errorMessage}.
     ${errorDetails ? `Error details: ${errorDetails}` : ""}`,
   });
-  return res.status(httpStatus).json({ error: errorMessage });
+  if (!res.writableEnded) {
+    return res.status(httpStatus).json({ error: errorMessage });
+  }
```
mongodben, line 58: what happens if there's an error and the writeable has ended? 

also are you sure this behavior holds for when the server isn't streaming responses?

packages/mongodb-rag-core/src/DataStreamer.ts
nlarew: `disconnect()` is a wrapper around `res.end()` which completes the request on the client.

This is actually a bit of a bug - if we get a `close` event then the client has already disconnected so it's too late to end the request.

packages/mongodb-chatbot-server/src/utils.ts
nlarew: This case probably _shouldn't_ happen if we structure things right, but it's a defensive check. If `writableEnded` is `true` then we've already finished responding and can't send anything else - the `.json()` call will throw an error.

PR #593, opened 2025-01-15
(EAI-625): LLM-as-a-judge evals on server messages

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-625

## Changes

- Add LLM-as-a-judge evals to messages.
  - Added to all messages in dev/staging. Only 10% in prod. 
- Add back ConversationAnswerRelevancy eval to `conversations.eval.ts` b/c upstream bug fixed.

## Notes

-


packages/chatbot-server-mongodb-public/src/tracing.ts
```diff
@@ -1,28 +1,46 @@
...
+        ...getTracingScores(tracingData, k),
+        ...(shouldJudge
+          ? await getLlmAsAJudgeScores(llmAsAJudge, tracingData)
+          : undefined),
       },
     });
   };
+};
+
+function validatePercentToJudge(percentToJudge?: number) {
```
nlarew, line 36: What if we combined this validator and the random number generation into a single function? Right now we're validating in two places (this function + `typeof llmAsAJudge?.percentToJudge === "number"` later)

e.g.

```ts
function shouldSample(percent: number | undefined): boolean {
  if (percentToJudge === undefined) {
    return false;
  }
  assert(
    percentToJudge >= 0 && percentToJudge <= 1,
    `percentToJudge must be between 0 and 1. Received: ${percentToJudge}`,
  );
  return Math.random() < percentToJudge;
}
```

packages/chatbot-server-mongodb-public/src/tracing.ts
```diff
@@ -76,6 +94,128 @@ export function extractTracingData(messages: SomeMessage[]) {
...
+    !tracingData.latestAssistantMessage
+  ) {
+    return;
+  }
+  const input = tracingData.latestUserMessage.content;
+  const output = tracingData.latestAssistantMessage.content;
+  const context = tracingData.latestUserMessage.contextContent
+    ?.map((c) => c.text)
+    // Have to do this type casting to make it compile. Unclear why.
+    .filter((c) => typeof c === "string") as string[] | undefined;
```
nlarew, line 162:
```suggestion
    .filter((c): c is string => typeof c === "string");
```

I think using a type predicate (i.e. the `is` keyword) should fix it without a cast

nlarew (approved):
LGTM % couple small suggestions

packages/chatbot-server-mongodb-public/src/tracing.ts
mongodben: ok, i def still dont grok `is`, but will see if this gets it

packages/chatbot-server-mongodb-public/src/tracing.ts
mongodben: more elegant will do

packages/chatbot-server-mongodb-public/src/tracing.ts
nlarew: tl;dr - `is` lets you return a `boolean` that also conveys type information.

This is necessary to get types out of `Array.prototype.filter` because JS is a crazy language.

PR #594, opened 2025-01-15
(EAI-653): Run LLM-as-a-judge evals on all rated messages

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-653

## Changes

- Run LLM-as-a-judge evals on all rated messages
- Standardize updateTracing func for all routes

## Notes

- Note: there'll be a bit of re-running LLM-as-a-judge evals on rated messages where a message is already LLM as a judge evaluated. Given that approx 2% of messages are rated, and we're running evals on on 10% of all messages. This means that evals will be rerun on  .2% of all messages (.1 * .02).


nlarew (approved):
LGTM

re: the re-run evals, doesn't seem like a big deal ultimately. That said, should we store info about evals in the `conversations` collection? e.g. a flag (e.g. `Message.hasEval`) or pointers to the evals in braintrust (e.g. `Message.evals = ["<link to braintrust>"]`?

mongodben:
> re: the re-run evals, doesn't seem like a big deal ultimately. That said, should we store info about evals in the `conversations` collection? e.g. a flag (e.g. `Message.hasEval`) or pointers to the evals in braintrust (e.g. `Message.evals = ["<link to braintrust>"]`?

thats an interesting idea but i'd rather not so that the tracing/online evals can be fully non-mutative, just exists on top of the existing behavior

PR #602, opened 2025-01-28
Truncate embeddings

mongodben:
Jira: n/a
## Changes

- Truncate the embeddings in the docs so that all the numbers in the vector aren't present. 

## Notes

- This will reduce the number of embeddings to make by multiple thousands
- Also allows us to load whole pages into context window where wouldnt have been able to b/c of all the extra embedding tokens


packages/ingest-mongodb-public/src/sources/snooty/truncateEmbeddings.ts
```diff
@@ -0,0 +1,8 @@
+export function truncateEmbeddings(body: string): string {
+  const embeddingRegex =
+    /([-]?0\.\d+d?),\s*([+\-]?0\.\d+d?)(?:\s*,\s*[+\-]?0\.\d+d?)*/g;
```
cbush, line 3: comment explaining what this is "supposed" to do?

packages/ingest-mongodb-public/src/sources/snooty/truncateEmbeddings.ts
```diff
@@ -0,0 +1,9 @@
+export function truncateEmbeddings(body: string): string {
+  //Matches embedding vector strings in the format: "0.123, 0.456, -0.789, ..."
```
nlarew, line 2:
```suggestion
  // Matches elements of an embedding vector string with at least two elements.
  // Each element must be between -1 and 1, include the leading 0 or -0, and
  // include at least one digit after the decimal point.
  // e.g. "0.123, 0.456, -0.789, ..."
```

packages/ingest-mongodb-public/src/sources/snooty/SnootyDataSource.ts
```diff
@@ -383,7 +384,7 @@ export const handlePage = async (
     ),
     sourceName,
     title,
-    body,
+    body: truncateEmbeddings(body),
```
nlarew, line 387: I don't love that this blanket applies to every page in the docs, though it's likely fine in practice since we don't directly show context to users and LLMs can probably infer the truncated bits.

I could imagine e.g. an aggregation operator example that happens to use an array of floats that meets the truncation criteria but isn't explicitly a vector embedding.

nlarew (approved):
LGTM - function seems to work, though I have a very minor concern about it being too broadly applied

packages/ingest-mongodb-public/src/sources/snooty/SnootyDataSource.ts
mongodben: TIL renard series...yea i think that's a valid concern. and certainly would be more of a concern if human readers were the main consumers. 

i think fine as is for now but something to consider if the MD is reappropriated for human readers

PR #603, opened 2025-01-28
(EAI-571) [UI] Support creating conversation on first message

nlarew:
Jira: (EAI-571) [UI] Support creating conversation on first message

## Changes

- No more calls to `createConversation` - instead we always create the conversation on the first message
  - Works locally with both streaming and awaited responses

## Notes

- We never really counted "raw interactions" anyway, but this notably will make it so our db doesn't track if someone opens that chatbot but doesn't send a message.

### DIFF
--- a/packages/mongodb-chatbot-ui/src/conversationStore.ts
+++ b/packages/mongodb-chatbot-ui/src/conversationStore.ts
@@ -6,7 +6,7 @@ import {
 import { combine } from "zustand/middleware";
 import createMessage, { CreateMessageArgs } from "./createMessage";
 import { References } from "./references";
-import { removeArrayElementAt } from "./utils";
+import { nonProd, removeArrayElementAt } from "./utils";
 
 export type ConversationState = {
   conversationId?: string;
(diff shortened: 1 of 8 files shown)


packages/mongodb-chatbot-ui/src/conversationStore.ts
```diff
@@ -49,12 +49,28 @@ export type ConversationStore = ReturnType<typeof makeConversationStore>;
...
-    combine(initialConversationState, (set) => ({
+    combine(initialConversationState, (set, get) => ({
       name,
       api: {
         initialize: (initialState: ConversationState) => {
           set(initialState);
         },
+        setConversationId: (conversationId: string) => {
+          if (conversationId === "") {
+            console.warn("Attempted to set an empty conversation ID");
```
mongodben, line 60: should these console.warn() calls be turned off for prod?

packages/mongodb-chatbot-ui/src/useConversation.tsx
mongodben, line 1: should there be some test for the new `"null"` behavior?

mongodben (approved):
lgtm, see comments.

works on my laptop

packages/mongodb-chatbot-ui/src/conversationStore.ts
nlarew: Probably doesn't matter that much, but for consistency with other warnings I'll update

packages/mongodb-chatbot-ui/src/useConversation.tsx
nlarew: Good call - added tests for streaming/awaited

PR #605, opened 2025-01-29
(EAI-655) fix mongo-memory-server test flakiness

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-655

## Changes

- instances of mongo memory server and mongo memory replica set are created in jest globalSetup, and stopped in globalTeardown instead of beforeAll and afterAll hooks in individual test files

## Notes

- globalSetup.ts is run only once. The server instance and replica set are created there with known server port and ip. The uri needed to connect to these can be determined by the ip and port. The uri is used in test files to connect to whichever the test file needs (server instance or replica set).

### DIFF
--- a/packages/benchmarks/jest.config.cjs
+++ b/packages/benchmarks/jest.config.cjs
@@ -3,4 +3,6 @@ module.exports = {
   testEnvironment: "node",
   testPathIgnorePatterns: ["<rootDir>/build"],
   verbose: true,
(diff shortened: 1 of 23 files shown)


mongodben (changes requested):
this looks good, but based on a quick look at the repo, it looks like `mongodb-chatbot-server` or `chatbot-server-mongodb-public` both use mongodb-memory-server. shouldn't those use the same pattern too?

mongodben (approved):

PR #606, opened 2025-01-30
(EAI-393) [UI] Chat window scrolls to bottom on copy button hover

nlarew:
Jira: (EAI-393) [UI] Chat window scrolls to bottom on copy button hover

## Changes

- Updates `@lg-chat/message-feed` to use the new "scroll to latest" button, which obviates the bug in this ticket.
- Updates all other LG dependencies to the latest version (otherwise npm is unhappy + dark mode breaks)

## Notes

-


mongodben (approved):
trusting this works, seems reasonable

PR #611, opened 2025-02-05
(EAI-639) setup WebDataSource

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-639

## Changes

- data source created for mongodb.com content
- processing sitemap to get urls for directories

## Notes

-


.drone.yml
```diff
@@ -14,6 +14,16 @@ trigger:
...
 steps:
+  - name: install_dependencies
+    image: node:18
+    commands:
+      - apt-get update
+      - apt-get install -yq gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 libexpat1 libfontconfig1 libgbm1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 ca-certificates fonts-liberation libnss3 lsb-release xdg-utils wget
+      - wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
+      - sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
+      - apt-get update
+      - apt-get install -y google-chrome-stable
```
mongodben, line 25: woah, i wonder if there's a clear way of doing this. i would imagine that there's a more straightforward way to do puppeteer in a linux box, but certainly could be wrong

packages/ingest-mongodb-public/package.json
```diff
@@ -48,16 +49,23 @@
     "typescript": "^5"
   },
   "dependencies": {
+    "@mozilla/readability": "^0.5.0",
```
mongodben, line 52: note: check to make sure you're using all the dependencies here. there might be some from when i was hacking around in the demo PR

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
```diff
@@ -0,0 +1,352 @@
...
+import xml2js from "xml2js";
+
+// TODO: move to constants file
+const sitemapURL = "https://www.mongodb.com/sitemap-pages.xml";
+// urls of the directories where we want every single page
+const directoryUrls = [
+  "https://www.mongodb.com/solutions/customer-case-studies",
+  "https://www.mongodb.com/solutions/solutions-library/",
+];
+const individualUrls = [
```
mongodben, line 16: all of this implementation specific detail should be passed as config to `makeWebDataSource()`

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
mongodben, line 1: nit: rename `MongoDbDotComWebDataSource` or something like that since this isn't generalizable to all websites

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
```diff
@@ -0,0 +1,352 @@
...
+            })
+          );
+        }
+        return pages;
+      } finally {
+        await browser?.close();
+      }
+    },
+  };
+}
```
mongodben, line 159:
```suggestion

interface MongoDbDotComWebDataSourceParams {
  name: string;
  urls?: [];
  sitemapPageUrls?: {
    sitemapUrl: string;
    // function to select only relevant pages from the sitemap entries.
    // note: could use regex but this is probably easier/more flexible
    filter: (sitemapUrls: string[]) => string[];
  }[]
}

export function makeWebDataSource({
  name,
  urls,
  sitemapPageUrls
}: MongoDbDotComWebDataSourceParams): DataSource {
  return {
    name,
    async fetchPages() {
      let browser;
      try {
        const { page: puppeteerPage, browser: b } = await makePuppeteer();
        browser = b;
        // TODO: add assertion to check that at least one of `urls` and `sitemapPageUrls` are defined
        
        // TODO: refactor the population of `pages` per the fact that urls/sitemapPageUrls can be passed in conditionally.
        // Use the sitemap to get the urls belonging to the directories we want every page from
        const completeDirectoryUrls = await getCompleteDirectoriesFromSitemap(
          sitemapURL
        );
        // Get urls from the list provided by web team
        const urls = [...completeDirectoryUrls, ...individualUrls];
        const pages: Page[] = [];
        for await (const [index, url] of urls.entries()) {
          if (maxPages && index >= maxPages) {
            break;
          }
          pages.push(
            await scrapePage({
              url,
              puppeteerPage,
            })
          );
        }
        return pages;
      } finally {
        await browser?.close();
      }
    },
  };
}
```

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
```diff
@@ -0,0 +1,352 @@
...
+  sitemapURL: string
+): Promise<string[]> {
+  const response = await fetch(sitemapURL);
+  const sitemap = await response.text();
+  const parser = new xml2js.Parser();
+  const parsedXML = await parser.parseStringPromise(sitemap);
+  return parsedXML.urlset.url.map((url: { loc: string[] }) => url.loc[0]);
+}
+
+export async function getCompleteDirectoriesFromSitemap(
```
mongodben, line 171:
```suggestion
async function getCompleteDirectoriesFromSitemap(
```

you dont use this func outside the test, dont export

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
```diff
@@ -0,0 +1,352 @@
...
+  puppeteerPage,
+  url,
+}: {
+  puppeteerPage: PuppeteerPage;
+  url: string;
+}): Promise<Page> {
+  // TODO: when productionizing, don't re-instantiate the browser on every call
+
+  try {
+    await puppeteerPage.goto(url, {
```
mongodben, line 336: i think i forgot to mention this before or in the ticket, but pls add the query parameter `?optimizely_x=0` to all requests to the page from puppeteer. this blocks the page from including optimizely experiments. 

only include the param here when calling puppeteer, not in the ingested page url.

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
```diff
@@ -0,0 +1,352 @@
...
+    console.log(`failed to open the page: ${url} with the error: ${error}`);
+  }
+
+  const content = await getContent(puppeteerPage);
+
+  return {
+    url,
+    format: "md",
+    // TODO: hoist to dataSource param
+    sourceName: "mongodb-dot-com",
```
mongodben, line 349: def want to pass this as a param

mongodben (changes requested):
nice start, see changes

packages/ingest-mongodb-public/src/sources/WebDataSource.ts
yakubova92: MongoDbDotComDataSource already exists. There's a task to deprecate it in this epic. Ok to rename after deprecating to avoid confusion?

packages/ingest-mongodb-public/package.json
mongodben, line 1: just checking, do you use all of these? i think so but want to make sure there isn't anything unused from the hacking i was doing

packages/ingest-mongodb-public/src/sources/mongodb-dot-com/WebDataSource.test.ts
```diff
@@ -0,0 +1,147 @@
...
+import {
+  getUrlsFromSitemap,
+  prepareWebSources,
+  RawWebSource,
+} from "./webSources";
+import fs from "fs";
+import path from "path";
+jest.setTimeout(60000);
+
+global.fetch = jest.fn();
```
mongodben, line 12: can this be scoped to the `getUrlsFromSitemap` describe block?

packages/ingest-mongodb-public/package.json
yakubova92: I removed @mozilla/readability and jsdom

packages/ingest-mongodb-public/src/sources/mongodb-dot-com/webSources.ts
```diff
@@ -0,0 +1,234 @@
+import puppeteer from "puppeteer";
+import xml2js from "xml2js";
+
+export type RawWebSource = {
```
mongodben, line 4: nit: i dislike the word 'raw' here. feels harsh/negative. can you consider different naming convention like `InitialWebSource`,  `UnprocessedWebSource`, etc.

packages/ingest-mongodb-public/src/sources/mongodb-dot-com/WebDataSource.test.ts
mongodben, line 1: for the director `mongodb-dot-com`, please rename to `mongodbDotCom` since we use camel case naming convention on directories w/in a package

packages/ingest-mongodb-public/package.json
yakubova92: xml2js is used to parse the sitemap, cheerio is used to pull elements from the html like title and meta tags, turndown is used to generate markdown

mongodben (changes requested):
generally looks good w/ a few small things. one major remaining thing like we talked about is the executablePath for puppeteer

packages/ingest-mongodb-public/src/sources/mongodb-dot-com/WebDataSource.ts
```diff
@@ -0,0 +1,210 @@
...
+  url,
+}: {
+  puppeteerPage: PuppeteerPage;
+  url: string;
+}): Promise<{ content: PageContent | null; error: string | null }> {
+  // TODO: when productionizing, don't re-instantiate the browser on every call
+  let content: PageContent | null = null;
+  let error: string | null = null;
+  try {
+    const response = await puppeteerPage.goto(`${url}?optimizely_x=0`, {
```
mongodben, line 199: rather than string interpolation, can you use the `URL` module to add the optimizely query param? bit safer

packages/ingest-mongodb-public/src/sources/mongodbDotCom/WebDataSource.ts
```diff
@@ -0,0 +1,220 @@
...
+type PageContent = Pick<Page, "body" | "metadata" | "title">;
+
+async function scrapePage({
+  browserPage,
+  url,
+}: {
+  browserPage: PlaywrightPage;
+  url: string;
+}): Promise<{ content: PageContent | null; error: string | null }> {
+  logger.info(`Scraping page: ${url}`);
```
yakubova92, line 203: Should we keep this logger?

nlarew: Might be better to move this up to line 30 of this file, i.e. in the for loop of `makeWebDataSource.fetchPages()`. That way we can keep the log in our ingest system but test the function / use it elsewhere without always logging.

packages/ingest-mongodb-public/package.json
```diff
@@ -58,6 +62,9 @@
     "mongodb-rag-ingest": "*",
     "node-fetch": "^2.7.0",
     "striptags": "^3.2.0",
+    "turndown": "^7.2.0",
+    "turndown-plugin-gfm": "^1.0.2",
+    "xml2js": "^0.6.2",
     "yaml": "^2.6.0"
   }
-}
+}
```
mongodben, line 70:
```suggestion
}

```

keep the newline at end pls

packages/ingest-mongodb-public/src/sources/mongodbDotCom/WebDataSource.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+    expect(webSources.length).toBe(2);
+    const [directoryWebSource, urlWebSource] = webSources;
+    expect(directoryWebSource.urls.length).toBe(4);
+    expect(directoryWebSource.urls).toStrictEqual(mockSitemapUrls);
+    expect(urlWebSource.urls).toStrictEqual(mockInitialWebSources[1].urls);
+  });
+});
+
+describe("WebDataSource", () => {
```
mongodben, line 83: pls add test for non-200 status page

packages/ingest-mongodb-public/src/sources/mongodbDotCom/webSources.ts
```diff
@@ -0,0 +1,233 @@
...
+    name: "web-misc",
+    urls: [
+      "https://www.mongodb.com",
+      "https://www.mongodb.com/atlas",
+      "https://www.mongodb.com/try/download/community",
+      "https://www.mongodb.com/pricing",
+    ],
+    staticMetadata: {
+      type: "Miscellaneous Web",
+    },
```
mongodben, line 179:
```suggestion
```

can remove since this doesn't provide real semantic value to the vector search where it's consumed

packages/ingest-mongodb-public/src/sources/mongodbDotCom/WebDataSource.test.ts
yakubova92: There is a test for this, it("handles invalid urls"). I just added a describe block for scrapePage, which is has more detailed tests for invalid urls

packages/ingest-mongodb-public/src/sources/mongodbDotCom/WebDataSource.ts
yakubova92: I can move this up and set it to logger.debug because its a lot of logs otherwise

PR #616, opened 2025-02-12
(EAI-868 & EAI-869): HuggingFace uploader & upload the docs dataset to HuggingFace 

mongodben:
Jira:
- https://jira.mongodb.org/browse/EAI-868
- https://jira.mongodb.org/browse/EAI-869

## Changes

- Create `uploadDatasetToHuggingFace()` to upload datasets to HuggingFace
- Create executable to upload the docs dataset to huggingface 
- Refactor repo to not be a JS module. Needed to do this to make test suite run. Also puts it in the same format as the rest of the packages in the monorepo. 

## Notes

- Used https://huggingface.co/datasets/mongodb-eai/test_integration for testing. You can see in there that this worked.


packages/datasets/src/bin/huggingFace/uploadContentDatasetToHuggingFace.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+  forbiddenUrls,
+  publicDatasetSourceName,
+} from "../../mongoDbDatasetConstants";
+import { uploadDatasetToHuggingFace } from "../../uploadDatasetToHuggingFace";
+import { HUGGINGFACE } from "../../EnvVars";
+import path from "path";
+
+async function uploadContentDatasetToHuggingFace() {
+  logger.info("Staring upload content dataset to Hugging Face script");
```
nlarew, line 17:
```suggestion
  logger.info("Starting upload content dataset to Hugging Face script");
```

packages/datasets/src/loadPageDataset.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+  @param pageStore - datastore for pages
+  @param dataSourceRegex - regular expression to filter pages by dataSource
+  @param forbiddenUrls - set of urls to exclude from the dataset
+ */
+export async function loadPagesDataset(
+  pageStore: MongoDbPageStore,
+  dataSourceRegex: RegExp,
+  forbiddenUrls: string[]
+): Promise<PageDatasetEntry[]> {
```
nlarew, line 17: Could we switch this to use an args object w/ corresponding type instead of positional args?

packages/datasets/src/uploadDatasetToHuggingFace.ts
```diff
@@ -0,0 +1,41 @@
+import { uploadFiles } from "@huggingface/hub";
+import { logger } from "mongodb-rag-core";
+
+export interface UploadDatasetToHuggingFaceParams {
+  hf: {
```
nlarew, line 5: Any particular reason to use `hf` instead of `huggingface`? Seems like we generally bias to more verbose names.

(if we were writing Swift we could use `🤗` as the field name!)

nlarew (approved):
LGTM - a few nits but nothing blocking

packages/datasets/src/EnvVars.ts
```diff
@@ -0,0 +1,11 @@
+export const OPENAI_ENV_VARS = {
+  OPENAI_API_KEY: "",
+  OPENAI_ENDPOINT: "",
+  OPENAI_CHAT_COMPLETION_DEPLOYMENT: "",
+  OPENAI_API_VERSION: "",
+};
```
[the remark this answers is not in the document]
mongodben: i think it was for specificity? not sure. this prob vestigate of hacking around. will clean up to use CORE_OPENAI_CONNECTION_ENV_VARS unless there's a clear reason that comes up not to do this.

packages/datasets/src/uploadDatasetToHuggingFace.ts
mongodben: not that i can recall. def prefer verbosity generally. will update.

PR #625, opened 2025-02-21
(EAI-642) replace static markdown files with web scraping

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-642

## Changes

- additional urls added to `web-misc` to replace markdown files from mongodb-corp directory
- all but one (overview.md) markdown file in mongodb-corp directory deleted
- documentation updated

## Notes

- It looks like most pages in the `mongodb-corp` data source can be handled by scraping a url rather than processing the static markdown file. There seems to be one clear exception, and a few cases where there is not an exact replacement. 

Exception:
overview.md describes the Chatbot capabilities and how it is built. The url listed in the markdown file is mongodb.com/docs, but this page does not have this information so scraping this site will not replace the overview.md file. Leaving this in and preserving mongoDbCorpDataSource. 

Murky cases:
1. mission.md, story.md, values.md were separate documents in the pages collection, all pointing to the mongodb.com/company url. This ticket would replace that and create one document for the mongodb.com/company url
2. leadership pages were separate documents in the pages collection, each pointing to a different fragment on the same url (eg: mongodb.com/leadership#<first-last-role>). This ticket would replace that and create one document for the mongodb.com/leadership role.

- This code is ready for review, but can only be merged to main after #611 is merged. The base branch is currently EAI-639 for easy comparison, but will be changed to main once #611 is merged


packages/ingest-mongodb-public/src/sources/mongodbDotCom/webSources.ts
```diff
@@ -173,6 +173,13 @@ export const initialWebSources: InitialWebSource[] = [
       "https://www.mongodb.com/atlas",
       "https://www.mongodb.com/try/download/community",
       "https://www.mongodb.com/pricing",
+      "https://www.mongodb.com/leadership",
+      "https://www.mongodb.com/legal/acceptable-use-policy",
+      "https://www.mongodb.com/legal/privacy-policy",
+      "https://www.mongodb.com/legal/terms-of-use",
+      "https://www.mongodb.com/why-use-mongodb",
+      "https://learn.mongodb.com",
+      "https://support.mongodb.com/",
```
mongodben, line 182: confirming that you checked that the ingestion for these pages actually works well. 

just run ingestion into dev and sanity check the outputs.

packages/ingest-mongodb-public/src/sources/mongodbDotCom/webSources.ts
```diff
@@ -173,6 +173,13 @@ export const initialWebSources: InitialWebSource[] = [
...
     ],
     staticMetadata: {
       type: "Miscellaneous Web",
```
mongodben, line 185: i think you can remove this type since it doesn't provide semantic value to vector search

yakubova92: removed this in EAI-639, where it was added

PR #629, opened 2025-02-24
(EAI-857) Create rST meta description import tool

nlarew:
Jira: (EAI-857) Create rST meta description import tool

## Changes

- Adds two new files (plus tests) to the `scripts` package
  - `meta2rst` contains the implementations that process and update reStructuredText files. The test file is fairly comprehensive and probably easier to review than the arcane implementations.
  - `upsertMetaDirective` which wraps the `meta2rst` functions in a script that we can call from the command line. We make this available via a new package.json script, `npm run upsertMetaDirective -- <path-to-rst-file> <meta-description>`. I tested this manually against some files in `docs-realm` and it seems to work perfectly. I'll probably wrap it up in a `.sh` file for the actual imports so that we can call it multiple time programmatically.

### DIFF
--- a/packages/scripts/src/meta2rst.ts
+++ b/packages/scripts/src/meta2rst.ts
@@ -1,3 +1,5 @@
+import fs from "fs";
+
 // Match the entire meta directive block
 const metaDirectiveRegex = /(?:^|\n)(.. meta::\s*\n(?:\s+:[^:]+:[^\n]*\n)*)/g;
 // To find specific fields within the matched directive
@@ -13,6 +15,7 @@ export function updateMetaDescription(
   newDescription: string
 ): string {
   return content.replace(metaDirectiveRegex, (match) => {
+    console.log("meta directive match", match);
     // Check if description field exists
     if (descriptionFieldRegex.test(match)) {
       // Replace existing description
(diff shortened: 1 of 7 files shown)


packages/scripts/src/meta2rst.ts
```diff
@@ -0,0 +1,176 @@
...
+  keywords: string | null;
+}): string {
+  if (args.description === null && args.keywords === null) {
+    return "";
+  }
+
+  const metaDirectiveParts = [`.. meta::`];
+
+  if (args.keywords) {
+    metaDirectiveParts.push(`   :keywords: ${args.keywords}`);
```
mongodben, line 57: 3 spaces!

packages/scripts/src/meta2rst.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (pageTitleLineNumber === -1) {
+      throw new Error("Page title not found");
+    }
+    const pageLines = rstPageContent.split("\n");
+    const newRstPageContent = [
+      ...pageLines.slice(0, pageTitleLineNumber),
+      "",
+      metaDirective,
+      "",
+      ...pageLines.slice(pageTitleLineNumber),
```
mongodben, line 157: are there some scenarios where there could only be 1 newline surrounding the metadirective, not 2?

packages/scripts/src/upsertMetaDirective.ts
```diff
@@ -0,0 +1,33 @@
...
+const metaDescription = args[1];
+if (!metaDescription) {
+  console.error("Meta description is required");
+  process.exit(1);
+}
+
+try {
+  upsertMetaDirectiveInFile(filePath, {
+    description: metaDescription,
+    keywords: null,
```
mongodben, line 26: why always no keywords?

packages/scripts/src/upsertMetaDirective.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+  console.log(`Updated meta description in ${filePath}`);
```
mongodben, line 28: use logger from mongodb-rag-core?

mongodben (approved):
great work here! 

couple kvetches about rst + a couple small questions

packages/scripts/src/meta2rst.ts
nlarew: It might work? Never really tried / paid attention to that case. Either way, it's always valid to have an empty line before and after the directive so this shouldn't be an issue.

packages/scripts/src/upsertMetaDirective.ts
nlarew: Just because we don't do any generative stuff for that currently. Passing `null` here is a no-op as opposed to an overwrite - it'll leave any existing keywords alone.

PR #634, opened 2025-02-27
[benchmarks] Use Braintrust proxy for Azure OpenAI and AWS Bedrock

mongodben:
Jira: n/a

## Changes

- For benchmarks, run AWS and Azure OpenAI models thru Braintrust proxy
- In benchmarks package remove following model calling logic:
  - Radiant pass-through (we haven't used in a while)
  - AWS Bedrock (now calling thru Braintrust)
  - Azure OpenAI (now calling thru Braintrust)
- Removed now-used logic for creating fake system prompts w/in a user message.
  - This used to be needed for some providers (forget which), but no longer is.
 
## Notes

- Cannot run Gemini models through proxy b/c we don't have permission to create a machine access token. We had been previously told that this would be hard/impossible to get. Not worth pursuing imo.


packages/benchmarks/src/models.ts
```diff
@@ -115,116 +98,111 @@ export const models: ModelConfig[] = [
...
     label: "nova-pro-v1:0",
     deployment: "amazon.nova-pro-v1:0",
     developer: "Amazon",
-    provider: "aws_bedrock",
+    provider: "braintrust",
     maxConcurrency: 1,
     authorized: true,
   },
-  // Note: not available in Bedrock (yet?)
+  // TODO: not working
```
mongodben, line 192:
```suggestion
```

packages/benchmarks/.env.example
```diff
@@ -8,17 +8,8 @@ MONGODB_TEXT_TO_DRIVER_CONNECTION_URI="<CONNECTION URI>"
...
 BRAINTRUST_API_KEY="<some api key>"
 BRAINTRUST_TEXT_TO_DRIVER_PROJECT_NAME="<some project>"
-# Note: We use the Azure-OpenAI style passthrough for the Radiant endpoint
-RADIANT_ENDPOINT="<base_url>/azure"
-RADIANT_API_KEY="rad-API_KEY-iant"
 OPENAI_API_VERSION="2024-06-01"
 OPENAI_ENDPOINT="https://<resource_name>.openai.azure.com/"
 OPENAI_API_KEY="<api_key>"
-# Cookie must be updated daily
-MONGODB_AUTH_COOKIE="<Auth cookie from CorpSecure-protected site>"
```
nlarew, line 18: Glad to see this particular piece of jank get removed

packages/benchmarks/src/models.ts
```diff
@@ -6,12 +6,7 @@ export type ModelDeveloper =
   | "Mistral"
   | "Amazon";
 
-export type ModelProvider =
-  | "radiant"
-  | "braintrust"
-  | "azure_openai"
-  | "aws_bedrock"
-  | "gcp_vertex_ai";
+export type ModelProvider = "braintrust" | "azure_openai" | "gcp_vertex_ai";
```
nlarew, line 9: Not against keeping `azure_openai` as an option - just confirming that we won't actually be using it for now?

nlarew (approved):

packages/benchmarks/src/models.ts
mongodben: oh we can remove. adding to PR

```suggestion
export type ModelProvider = "braintrust" | "gcp_vertex_ai";
```

PR #648, opened 2025-03-07
(EAI-911) fix bug causeing BinaryNdcgAt5 errors

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-911

## Changes

- fix fuzzy match for case where one of the urls has no path. This was causing more matches than possible, leading to a score > 1, when the score must be between 0 and 1
- fixed an existing test case
- added an additional test case

## Notes

- tested this with some evals, not seeing BinaryNdcgAt5 errors anymore

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/eval/fuzzyLinkMatch.test.ts
+++ b/packages/chatbot-server-mongodb-public/src/eval/fuzzyLinkMatch.test.ts
@@ -64,7 +64,7 @@ describe("fuzzyLinkMatch", () => {
     expect(fuzzyLinkMatch(expected, actual)).toBe(false);
   });
 
-  test("actual url has not path, but expected does", () => {
+  test("expected url has path, but actual does not", () => {
     const expected = "https://example.com/path/to/resource";
     const actual = "https://example.com";
     expect(fuzzyLinkMatch(expected, actual)).toBe(false);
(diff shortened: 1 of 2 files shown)


packages/chatbot-server-mongodb-public/src/eval/fuzzyLinkMatch.test.ts
```diff
@@ -57,4 +57,16 @@ describe("fuzzyLinkMatch", () => {
...
     expect(fuzzyLinkMatch(expected, actual)).toBe(true);
   });
+
+  test("expected url has no path, but actual does", () => {
+    const expected = "https://example.com";
+    const actual = "https://example.com/path/to/resource";
+    expect(fuzzyLinkMatch(expected, actual)).toBe(false);
+  });
+
+  test("actual url has not path, but expected does", () => {
```
nlarew, line 67:
```suggestion
  test("expected url has path, but actual does not", () => {
```
Slight reword so that `expected` comes first like in the other tests

packages/chatbot-server-mongodb-public/src/eval/fuzzyLinkMatch.ts
```diff
@@ -1,20 +1,40 @@
...
+  const cleanExpectedPath = getCleanPath(expected);
 
-  If either input is not a valid URL,
-  it treats the input as a plain string
-  and performs a substring match.
+  // if cleaned path is not an empty string, compare cleaned paths
+  if (cleanActualPath && cleanExpectedPath) {
+    return cleanActualPath.endsWith(cleanExpectedPath);
+  } else {
+    // compare full normalized paths
```
nlarew, line 24:
```suggestion
    // compare normalized full URLs
```
I think this is more accurate? Disregard if I'm wrong.

nlarew (approved):
Thanks for cleaning this up! A couple nits then LGTM

PR #649, opened 2025-03-07
(EAI-523) Server supports clientContext passed in requests

nlarew:
Jira: (EAI-523) Server supports clientContext passed in requests

## Changes

- Adds a new, optional `clientContext` request body parameter for `addMessage` requests
- If defined, this value is passed as an arg to `generateUserPrompt`

## Notes

- This PR adds support in the framework but does not modify our server implementation. For that, we'll want to do some light eval work + handle specific data from clients.

### DIFF
--- a/packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
+++ b/packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
@@ -44,6 +44,7 @@ export const DEFAULT_MAX_USER_MESSAGES_IN_CONVERSATION = 7; // magic number for 
 export type AddMessageRequestBody = z.infer<typeof AddMessageRequestBody>;
 export const AddMessageRequestBody = z.object({
   message: z.string(),
+  clientContext: z.object({}).passthrough().optional(),
 });
 
 export const AddMessageRequest = SomeExpressRequest.merge(
@@ -103,6 +104,7 @@ export interface AddMessageToConversationRouteParams {
 type MakeTracedResponseParams = Pick<
   GenerateResponseParams,
   | "latestMessageText"
+  | "clientContext"
   | "customData"
   | "dataStreamer"
   | "shouldStream"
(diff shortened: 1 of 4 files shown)


mongodben:
> * This PR adds support in the framework but does not modify our server implementation. For that, we'll want to do some light eval work + handle specific data from clients.

can you create separate ticket for this?

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -44,6 +44,7 @@ export const DEFAULT_MAX_USER_MESSAGES_IN_CONVERSATION = 7; // magic number for
 export type AddMessageRequestBody = z.infer<typeof AddMessageRequestBody>;
 export const AddMessageRequestBody = z.object({
   message: z.string(),
+  clientContext: z.object({}).passthrough().optional(),
```
mongodben, line 47: what does `.passthrough()` do here?

mongodben, line 47: tangential not for this PR: i think it'd be nice if for these 'custom' types, we had them be passable zod types, so you can specify the shape for your implementation, and then it's automatically validated on server

packages/mongodb-chatbot-server/src/routes/generateResponse.test.ts
```diff
@@ -198,7 +198,7 @@ describe("generateResponse", () => {
       return {
         references,
         userMessage: {
-          role: "user",
+          role: "user" as const,
           content: userMessageText,
         },
```
mongodben, line 203: aside: i've always done `{ role: "user", content:  "..."} satisfies SomeType` to get around the type err here. good to kno this works as well

mongodben (approved):
lgtm, can do minimal docs update as well? namely the openapi spec

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
nlarew: Passthrough is basically equivalent to `additionalProperties: true` in JSON schema. If you don't have it then zod parsing strips out any fields that aren't in the schema. Might be unnecessary here if we go the custom schema route you suggested.

PR #652, opened 2025-03-10
chore: update dependencies

HuberTRoy:
Hi~, when i tried running this example by following this doc: https://mongodb.github.io/chatbot/quick-start/.

I got the error:

```
Error: Cannot find module 'jsdom'
Require stack:
...
```

It seems to have lost the required dependencies, after installed them, everything ran as the doc's said.

### DIFF
--- a/examples/quick-start/package-lock.json
+++ b/examples/quick-start/package-lock.json
@@ -12,6 +12,7 @@
         "packages/*"
       ],
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


mongodben:
thanks for bringing this up. we've made the decision to deprecate the framework and its examples. more info here: https://mongodb.github.io/chatbot/framework

closing this issue as it's no longer relevant to the direction of the project.

PR #657, opened 2025-03-12
(EAI-645) deprecate coachGTM

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-645

## Changes

- remove all references to coachGTM

## Notes

- the team that was using coachGTM (Tech Ops) will now be using the same database we use for all teams (web, docs, university)
- the coach-gtm db is still available on staging, but will no longer be updated due to these changes. Once the Tech Ops team has moved  off the coach-gtm db, we can removed that db


packages/chatbot-server-mongodb-public/src/eval/bin/urlCheck.ts
mongodben, line 1: can you add a @fileoverview comment at the top specifying what this code is for

packages/ingest-mongodb-public/src/coachGtm.config.ts
mongodben, line 1: pls also delete legacy `makeMongoDbDotComDataSource` from repo that uses the web cms

packages/chatbot-server-mongodb-public/src/eval/bin/urls.ts
mongodben, line 1: i dont love committing these files to repo. can you read from/write to the local file system instead of committing these?

mongodben (commented):
back to you w/ a few things. generally looks good tho

packages/chatbot-server-mongodb-public/src/eval/bin/urlCheck.ts
```diff
@@ -0,0 +1,85 @@
...
+      noRedirect.add(url);
+    }
+  }
+  return {
+    noRedirect,
+    redirectTo,
+  };
+};
+
+async function main() {
```
nlarew, line 59: This feels more like something that belongs in the `scripts` package than in our server. It's pretty specific to a one-time use case (e.g. "sergio" in one of the variable names). It also isn't doing evaluations so feels out of place in the `evals` directory.

yakubova92: I intended to remove this first commit from the PR, I just forgot about it. This was just to help me get a list of urls that sergio's team was using that aren't being ingested. Ben and I spoke about this recently and decided to keep the script since it may be useful in the future. I made some changes in my [last commit](https://github.com/mongodb/chatbot/pull/657/commits/5ab1cc783fa4dcb1340b346ea9ac8fddf1056166) to make this script more re-usable and generic. It now reads from and writes to the local file system.

yakubova92: I have moved this to the scripts directory

PR #661, opened 2025-03-17
Update stats materialized view script

nlarew:
This updates the code for our materialized view stats collection that we can use to get historical operational data (e.g. num messages, num conversations) and make charts without needing to re-calculate every time. Instead we _pre-_calculate nightly.


mongodben (approved):
lgtm but can you clarify in PR description why you're adding this code?

PR #675, opened 2025-04-16
include the authUser in slack msgs

mongodben:
Jira: n/a

## Changes

- When using chatbot behind the MDB CorpSecure (such as staging site https://chat-server.docs.staging.corp.mongodb.com/), we add the `auth_user` cookie to braintrust logs and also post to slack
  - this is useful for manual testing/feedback of chatbot

## Notes

- Note, this won't work w/ the vanity URL staging site https://knowledge.staging.corp.mongodb.com/ b/c this is only behind the firewall, not corpsecure.

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/tracing/postCommentToSlack.ts
+++ b/packages/chatbot-server-mongodb-public/src/tracing/postCommentToSlack.ts
@@ -149,22 +149,31 @@ ${
     : "No LLM-as-Judge Scores"
 }`;
 
+  const messageBlocks = [
+    Blocks.Section({ text: feedbackMd }),
+    Blocks.Divider(),
+    Blocks.Section({ text: messagesMd }),
+    Blocks.Section({ text: referencesMd }),
+    Blocks.Divider(),
+    Blocks.Section({ text: scoresMd }),
+    Blocks.Divider(),
+    Blocks.Section({ text: messageMetadataMd }),
+    Blocks.Divider(),
+    Blocks.Section({ text: idMetadataMd }),
+  ];
+
+  const maybeAuthUser = conversation.customData?.authUser;
+  if (maybeAuthUser && typeof maybeAuthUser === "string") {
+    messageBlocks.unshift(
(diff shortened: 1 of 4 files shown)


packages/chatbot-server-mongodb-public/src/tracing/postCommentToSlack.ts
```diff
@@ -149,22 +149,31 @@ ${
...
+    Blocks.Divider(),
+    Blocks.Section({ text: idMetadataMd }),
+  ];
+
+  const maybeAuthUser = conversation.customData?.authUser;
+  if (maybeAuthUser && typeof maybeAuthUser === "string") {
+    messageBlocks.unshift(
+      Blocks.Section({ text: `Auth User: ${Md.user(maybeAuthUser)}` })
+    );
+  }
```
nlarew, line 170: This looks pretty similar to the work we do in `routesUpdateTraceHandlers` - should we pass in `authUser` from there as an arg to this function instead of recomputing?

nlarew (approved):
LGTM % a nit

packages/chatbot-server-mongodb-public/src/tracing/routesUpdateTraceHandlers.ts
```diff
@@ -27,6 +27,11 @@ export const makeAddMessageToConversationUpdateTrace: (
       typeof llmAsAJudge?.percentToJudge === "number" &&
       Math.random() < llmAsAJudge.percentToJudge;
 
+    const maybeAuthUser = conversation.customData?.authUser;
+    if (maybeAuthUser && typeof maybeAuthUser === "string") {
```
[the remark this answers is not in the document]
mongodben: it'll always be a string (presumably) but it's not typed in the code so doing typecheck here

packages/chatbot-server-mongodb-public/src/tracing/postCommentToSlack.ts
mongodben: no i dont think so b/c we don't have a clean way to pass w/o major refactors

PR #679, opened 2025-04-19
(EAI-950) Segment - Send data feed to Segment

nlarew:
Jira: (EAI-950) Segment - Send data feed to Segment

## Changes

- Defines a new `useSegmentIds` middleware to pull out Segment user IDs from request headers
- Defines `segment` handlers that send chatbot events to Segment
  - Uses the `SEGMENT_WRITE_KEY` env var to route requests
- Updates post-request trace handlers to also send segment events


packages/chatbot-server-mongodb-public/src/config.ts
```diff
@@ -225,42 +228,69 @@ const llmAsAJudgeConfig = {
...
       : undefined,
+    addMessageToConversationCustomData: async (req, res) => {
+      const defaultCustomData = await defaultAddMessageToConversationCustomData(
+        req,
+        res
+      );
+      return {
+        ...defaultCustomData,
+        segmentUserId: res.locals.customData.segmentUserId ?? "NOPE",
+        segmentAnonymousId: res.locals.customData.segmentAnonymousId ?? "NOPE",
```
mongodben, line 258: can you do `null` rather than `'NOPE'`? or does 'NOPE' mean something to segment

packages/chatbot-server-mongodb-public/src/tracing/routesUpdateTraceHandlers.ts
```diff
@@ -32,6 +57,36 @@ export const makeAddMessageToConversationUpdateTrace: (
...
+        isVerifiedAnswer: tracingData.isVerifiedAnswer ?? false,
+        rejectionReason: tracingData.rejectQuery
+          ? (assistantMessage.customData?.rejectionReason as
+              | string
+              | undefined) ?? "Unknown rejection reason"
+          : undefined,
+      });
+    }
+
     logger.updateSpan({
```
mongodben, line 90: now that there are 2 separate things happening in this trace func, should we wrap each w/ a try/catch block? 

i know both aren't async, since the processing happening in background..but still. 

not sure what right approach is, but wanted to raise the question

packages/chatbot-server-mongodb-public/src/middleware/useSegmentIds.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function useSegmentIds(): ConversationsMiddleware {
+  return (req, res, next) => {
+    const userId = req.header(SEGMENT_USER_ID_HEADER);
+    const anonymousId = req.header(SEGMENT_ANONYMOUS_ID_HEADER);
+    if (userId) {
+      res.locals.customData.segmentUserId = userId;
+    }
+    if (anonymousId) {
+      res.locals.customData.segmentAnonymousId = anonymousId;
```
mongodben, line 14: by including this, does it mean that we're now tracking the segment user and anon IDs in the DB?

packages/chatbot-server-mongodb-public/src/config.ts
nlarew: Oops did that for testing purposes and used a silly word so that I'd notice it later. Good catch!

packages/chatbot-server-mongodb-public/src/middleware/useSegmentIds.ts
nlarew: Yes we will track these in user message `customData`. I think they classify as PII so don't want to keep them in `scrubbed_messages`.

nlarew:
> out of curiousity, whaat do the segment ID and anon IDs look like? and where are they coming from?

They're UUIDs with slightly different formats, e.g.

- `ajs_anonymous_id: "f32c22fe-510d-4d22-9122-5ea4af6b228d"`
- `ajs_user_id: "71ff22b6cf2cf2175312465f166a9a92f4f801fcef24d12c4aa79c65e8c20dab"`

They're stored in localStorage on MongoDB web properties that include Segment's AnalyticsJS (pretty much all of our sites, including docs)

> can we connect a segment ID back to a real user, like w/ email or smthn?

I think the `userId` can - will confirm with MOPS

PR #680, opened 2025-04-19
(EAI-949) Segment - Capture anonymousId and userId

nlarew:
Jira: (EAI-949) Segment - Capture anonymousId and userId

Related to https://github.com/mongodb/chatbot/pull/679

## Changes

- Exports a new function from the UI package that makes it easy to set the proper segment headers


packages/mongodb-chatbot-ui/src/App.tsx
```diff
@@ -84,7 +85,10 @@ function App() {
           serverBaseUrl={serverBaseUrl}
           shouldStream={shouldStream}
           darkMode={darkMode}
-          fetchOptions={{ credentials: "include" }}
+          fetchOptions={{
+            credentials: "include",
+            headers: getSegmentIdHeaders(),
```
mongodben, line 90: should this also be included in the docs/uni/devcenter implementations of the chatbot?

nlarew: They don't actually use the built in ones anymore - will make separate tickets for each team to integrate into their respective implementations.

PR #683, opened 2025-04-25
(EAI-968) ingest multiple versions

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-968

## Changes

- removed `snooty` prefix from sourceName
- ingesting all active branches for each data source
- inactive branches not ingested
- if pages exist for a branch that has recently become inactive, it will be deleted when ingest runs
- functionality to override what is considered the current version
- removed the override for docs v7.0 as [discussed in slack](https://mongodb.slack.com/archives/C05CE6UBGG5/p1745450845744029)
- updated tests

## Notes
- version is saved on Page on the metadata object. This is a deviation from the [spec doc](https://docs.google.com/document/d/1abwdBxOTxKLEynxSFeFFrrY-XGu4_UIL4XWq4fsVTnA/edit?usp=sharing), where we planned to set version on the root level of the document. I made this choice bc this is where version was previously stored, and only "snooty" pages are versioned.
- When this is approved, I will update the `pages_current` view in prod to filter out all but the current branch by adding the following the to the match stage
```
{
  $or: [
    {
      "metadata.version.isCurrent": true
    },
    {
      "metadata.version.isCurrent": {
        $exists: false
      }
    }
  ]
}
```
DO NOT MERGE to main until retrieval is set up to avoid non current versions. This PR will be merged to the feature branch

### DIFF
--- a/packages/ingest-mongodb-public/src/meta.config.ts
+++ b/packages/ingest-mongodb-public/src/meta.config.ts
@@ -71,7 +71,7 @@ export const standardConfig = {
   }),
   dataSources: async () => {
     const source = makeSnootyDataSource({
-      name: `snooty-${metaSnootyProject.name}`,
+      name: metaSnootyProject.name,
       project: {
         ...metaSnootyProject,
         currentBranch: metaSnootyProject.currentBranch,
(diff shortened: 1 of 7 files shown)


mongodben (approved):
everything looks good and makes sense to me.

nice work here

PR #687, opened 2025-04-28
(EAI-988): Refactor `GenerateResponse` for tool call support 

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-988

## Changes

- Simplify `GenerateResponse` so only takes the request-specific args
  - Can pass other stuff in a constructor function
- Refactor legacy RAG system to fit into shape of new construct.
  - To be removed relatively soon
- Include `GenerateResponse` in args to conversations server. 

## Notes

-


packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
```diff
@@ -31,13 +32,29 @@ import {
...
+  request?: ExpressRequest;
+}
+
+export interface GenerateResponseReturnValue {
+  messages: SomeMessage[];
+}
+
+export type GenerateResponse = (
+  params: GenerateResponseParams
+) => Promise<GenerateResponseReturnValue>;
```
nlarew, line 57: Should this all be in a separate `generateResponse.ts` file? Seems like a distinct (if not separate) concern from the route that calls it.

nlarew (approved):
LGTM - some minor comments + a typo that you might wanna fix

packages/mongodb-chatbot-server/src/processors/makeFilterNPreviousMessages.test.ts
```diff
@@ -21,25 +15,7 @@ const userMessage = {
 } satisfies Message;
 
 describe("makeFilterNPreviousMessages", () => {
-  it("should throw an error when there are no messages", async () => {
```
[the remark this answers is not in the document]
mongodben: correct. this is b/c we wont store the system message anymore. 

this is for a few reasons:
1. this is b.c some tool info is in the system msg. so if tool changes, then could be mismatch btwn behavvior and persisted system message
2. logically, the system msg isn't really part of the converastion, so probably should have never been saved 🤷

packages/mongodb-chatbot-server/src/routes/conversations/addMessageToConversation.ts
mongodben: yeah reasonable, will do

PR #690, opened 2025-04-29
Segment fixes

nlarew:
Jira: (EAI-997) [UI] Segment fixes

## Changes

- Strips literal quotes from segment ids (if present)
- Supports dynamic fetchOptions in conversations service
  - This lets us pull the segment headers on every request. Useful if they're not set until after the chatbot renders.

### DIFF
--- a/packages/mongodb-chatbot-ui/src/services/conversations.ts
+++ b/packages/mongodb-chatbot-ui/src/services/conversations.ts
@@ -153,10 +153,8 @@ const isConversationStreamEvent = (
  */
 export type ConversationFetchOptions = Omit<
   RequestInit,
-  "body" | "method" | "headers" | "signal"
-> & {
-  headers?: Headers;
-};
+  "body" | "method" | "signal"
+>;
 
 export type AddMessageRequestBody = {
   message: string;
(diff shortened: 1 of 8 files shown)


packages/mongodb-chatbot-ui/src/services/conversations.ts
```diff
@@ -165,12 +163,14 @@ export type AddMessageRequestBody = {
 
 export type ConversationServiceConfig = {
   serverUrl: string;
-  fetchOptions?: ConversationFetchOptions;
+  fetchOptions?: ConversationFetchOptions | (() => ConversationFetchOptions);
```
mongodben, line 166: dont love the overloading of this w/ 2 types as a design principle but meh

mongodben (approved):
lgtm, note the build err that must be fixed before merge. seems easy nuff to fix.

packages/mongodb-chatbot-ui/src/services/conversations.ts
nlarew: Yeah maybe not ideal but I've seen it before. Mainly here for backwards compatibility just in case.

PR #691, opened 2025-04-30
(EAI-998): Low-hanging improvements to existing tracing metrics

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-998

## Changes

- Add new metrics to conversation tracing
  - Track if message is rated (0 if not, 1 if so)
  - Track if message is commented (0 if not, 1 if so)
- Update existing metrics to compute across all traces
  - Redo `RejectedQuery` as InputGuardrailPass`. 1 if passes guardrail, 0 if not.
  - Redo `LlmDoesNotKnow` as `LlmAnswerAttempted`. 1 if attempts answer (not the canned idk response), 0 if canned answer
  - `VerifiedAnswer` now calculates 1 if verified answer, 0 if not. Previously was only calculated if it was a verified answer. 

## Notes

- This gives us better top-line numbers to view in braintrust logs, as the metrics are now across all messages, rather than just the ones where there was a pass.

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/tracing/routesUpdateTraceHandlers.ts
+++ b/packages/chatbot-server-mongodb-public/src/tracing/routesUpdateTraceHandlers.ts
@@ -126,9 +126,9 @@ function getTracingScores(
   k: number
 ) {
   return {
-    AcceptedQuery: tracingData.rejectQuery === true ? 0 : 1,
+    InputGuardrailPass: tracingData.rejectQuery === true ? 0 : 1,
     VerifiedAnswer: tracingData.isVerifiedAnswer === true ? 1 : 0,
-    LlmAnswered: tracingData.llmDoesNotKnow === true ? 0 : 1,
+    LlmAnswerAttempted: tracingData.llmDoesNotKnow === true ? 0 : 1,
     HasRating: tracingData.rating !== undefined ? 1 : 0,
     HasComment: tracingData.comment !== undefined ? 1 : 0,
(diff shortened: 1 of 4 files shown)


packages/chatbot-server-mongodb-public/src/tracing/routesUpdateTraceHandlers.ts
```diff
@@ -126,13 +126,25 @@ function getTracingScores(
...
 ) {
   return {
-    RejectedQuery: tracingData.rejectQuery === true ? 1 : null,
-    VerifiedAnswer: tracingData.isVerifiedAnswer === true ? 1 : null,
-    LlmDoesNotKnow: tracingData.llmDoesNotKnow === true ? 1 : null,
-    [`RetrievedChunksOver${k}`]:
-      tracingData.isVerifiedAnswer !== true
-        ? tracingData.numRetrievedChunks / k
-        : null,
+    // These metrics should start at 0,
```
nlarew, line 129: Not sure I follow what "start at 0" means - looks like we assign either 0 or 1 here vs assigning 0 and then later re-assigning 1.

nlarew:
LGTM as a concept - note the type errors in CI that need fixing

packages/chatbot-server-mongodb-public/src/tracing/routesUpdateTraceHandlers.ts
mongodben: practically speaking, it'll always be 0 since there's never a comment/rating when the msg is first sent.

PR #695, opened 2025-05-01
(EAI-1002): Dataset/benchmark results to huggingface

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-1002

## Changes

- Write up the huggingface readme w/ dataset overview + report
  - mostly just rehash of the report, but with some different points of emphasis
    - for example, removed reference to previous internal benchmarks/data, hoisted info about datasets to the top
- Also added the relevant dataset materials to huggingface https://huggingface.co/datasets/mongodb-eai/natural-language-to-mongosh
  - currently private til this is approved

## Notes

- when this is approved, i'll copy the .md file over to the huggingface readme and publish publicly


packages/ingest-mongodb-public/environments/production.yml
```diff
@@ -31,6 +31,34 @@ cronJobs:
         cpu: 4
         memory: 5Gi
     backoffLimit: 3
+  - name: ingest-pages-with-links
```
nlarew, line 34: This env update feels out of place - just marking here for you to merge or double check that it won't be an issue on merge

nlarew (approved):

packages/ingest-mongodb-public/environments/production.yml
mongodben: oh definitely. good catch. this was for other PR i was working on at similar time. must have gotten caught up in gitfoo.

will remove from PR

PR #703, opened 2025-05-07
(EAI-1001) enable search by type

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-1001

## Changes

- "metadata.pageType" field added
- SnootyDataSource creates pages with `"metadata.pageType": "tech-docs"`
- DevCenterDataSource creates pages with `"metadata.pageType": "devcenter"`
- index created for new field, vector search index updated to include new field

## Notes

- This will allow us to change the filter [here](https://github.com/mongodb/chatbot/blob/EAI-428-feature-versioned-docs/packages/datasets/src/mongoDbDatasetConstants.ts#L19-L22) to find public datasets (snooty or devcenter) by filtering on the "metadata.pageType" field

### DIFF
--- a/packages/mongodb-rag-core/src/contentStore/EmbeddedContent.ts
+++ b/packages/mongodb-rag-core/src/contentStore/EmbeddedContent.ts
@@ -38,6 +38,11 @@ export interface EmbeddedContent {
   updated: Date;
 
   /**
+    The source type indicates where the page was loaded from.
+   */
+  sourceType?: Page["sourceType"];
+
+  /**
     Arbitrary metadata associated with the content. If the content text has
     metadata in Front Matter format, this metadata should match that metadata.
    */
@@ -49,7 +54,6 @@ export interface EmbeddedContent {
       isCurrent: boolean;
       label: string;
     };
-    pageType?: string;
   };
 
   /**
@@ -89,7 +93,7 @@ export type QueryFilters = {
     current?: boolean;
     label?: string;
   };
-  pageType?: string;
+  sourceType?: string;
 };
 
 /**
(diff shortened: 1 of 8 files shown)


packages/mongodb-rag-core/src/contentStore/Page.ts
```diff
@@ -48,6 +48,12 @@ export type PageMetadata = {
     If the page is not versioned, this field should be undefined.
    */
   version?: VersionInfo;
+  /**
+    The page type. Other types may be added in the future. 
+    "tech-docs" indicates documents from the mongodb.com/docs site. SnootyDataSource has this type
+    "devcenter" indicates documents from the mongodb.com/developer site. DevCenterDataSource has this type
+   */
+  pageType?: "tech-docs" | "devcenter";
```
mongodben, line 56: a few things:

1. pls rename this to `contentType`. more descriptive
2. do not hard code in the values on the Page type. keep as `string`. remember, this is a flexible construct. also this type is used by pages from other sources, such as web data sources, etc.

mongodben: ah ok. in that case can you put the value inside of the Page.metadata.page object? this way its not included in the chunked data.

so the property would be `Page.metadata.page.type?: string`

packages/mongodb-rag-core/src/contentStore/EmbeddedContent.ts
```diff
@@ -88,6 +93,7 @@ export type QueryFilters = {
     current?: boolean;
     label?: string;
   };
+  sourceType?: string;
```
mongodben, line 96: shuld this also be Page["sourceType"] ?

mongodben (approved):
lgtm w/ a few very small things to address before merge

PR #776, opened 2025-06-12
EAI-1053 Validation on ingest: 1 page in each source

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1053

## Changes

- Cronjob that verifies all sources have at least 1 page (i.e. a source's entire content is not marked as "action": "deleted")


packages/scripts/src/verifyPagesInSource.ts
```diff
@@ -0,0 +1,132 @@
...
+*/
+import { assertEnvVars, makeMongoDbPageStore } from "mongodb-rag-core";
+import "dotenv/config";
+
+const { MONGODB_DATABASE_NAME, MONGODB_CONNECTION_URI } = assertEnvVars({
+  MONGODB_DATABASE_NAME: "",
+  MONGODB_CONNECTION_URI: "",
+});
+
+const excludeSources: Set<string> = new Set([
```
nlarew, line 13: Do we exclude these sources because we already have integration tests for them? Thoughts on unifying the approach here longer term?

packages/scripts/src/verifyPagesInSource.ts
```diff
@@ -0,0 +1,132 @@
...
+  "snooty-java-rs",
+  "mongodb-dot-com",
+  "web-company",
+  "web-customer-case-studies",
+  "web-solutions-library",
+  "web-university",
+  "web-services",
+]);
+
+async function main() {
```
nlarew, line 65: To check my understanding - this script finds sources that at one point existed in our collection but have since been deleted without any replacements from the same source.

I think this approach is valid just want to make sure that's what we're doing.

packages/scripts/src/verifyPagesInSource.ts
hschawe: also, long-term we should probably be cleaning removed or renamed sources from our DB

PR #782, opened 2025-06-20
chore: migrate type tests to TSTyche

mrazauskas:

## Changes

This PR migrates type test to TSTyche.

## Notes

`jest-tsd` was added recently to this repo. It depends on `tsd-lite` that is [deprecated for long time](https://github.com/mrazauskas/tsd-lite/issues/364). I was maintaining `tsd-lite` and I would recommend using TSTyche instead. This is my new project written from scratch. TSTyche solves many issues and shortcomings of `tsd` architecture.

For instance:

- it has `test()` and `describe()` with `.only` and `.skip`;
- `expect` style assertions that read just like any other test in this repo;
- assertions like `.toHaveProperty()` or `.toBeCallableWith()`, you don’t have to expect errors anymore;
- it can test agains specified version of TypeScript or a range of versions: `tstyche --target '>=5.0 <5.3'`;
- the install size is only 250kB, because it will load the installed `typescript`;
- and so on.

Check out the documentation: https://tstyche.org

### DIFF
--- a/packages/mongodb-rag-core/package.json
+++ b/packages/mongodb-rag-core/package.json
@@ -41,7 +41,7 @@
   "scripts": {
     "clean": "rm -rf build",
     "build": "tsc -b tsconfig.build.json",
-    "test": "jest --forceExit && npm run test:types",
+    "test": "jest --forceExit",
     "test:types": "tstyche",
     "lint": "eslint ./src --ext ts,tsx,js,jsx --report-unused-disable-directives",
     "lint:fix": "npm run lint -- --fix && prettier ./src --check --write",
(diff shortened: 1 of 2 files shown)


packages/mongodb-rag-core/package.json
```diff
@@ -32,7 +32,6 @@
     "./mongodb": "./build/mongodb.js",
     "./mongoDbMetadata": "./build/mongoDbMetadata/index.js",
     "./openai": "./build/openai.js",
-    "./aiSdk": "./build/aiSdk.js",
```
mrazauskas, line 35: Was removed by `npm` as duplicate. Already defined above (line 30).

PR #791, opened 2025-06-26
(EAI-1085) fetch_page tool

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1085

## Changes

- **adds new tool for fetching a single page based on URL**
- adds core URL normalization function (but doesn't normalize URLs during fetch_page yet since it's not in ingestion) 
- renamed generateResponseWithSearchTool.ts -> generateResponseWithTools.ts


packages/chatbot-server-mongodb-public/src/processors/formatUserMessageForGeneration.ts
```diff
@@ -0,0 +1,48 @@
+/**
+  Adds front matter to the user message text.
+ */
```
mongodben, line 3:
```suggestion
```

packages/chatbot-server-mongodb-public/src/processors/formatUserMessageForGeneration.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  if (originCode === "VSCODE") {
+    frontMatters.client = "MongoDB VS Code plugin";
+  } else if (originCode === "GEMINI_CODE_ASSIST") {
+    frontMatters.client = "Gemini Code Assist";
+  }
+  return addFrontMatter(frontMatters, userMessageText);
+}
+
+function addFrontMatter(
```
mongodben, line 35: there's a utility func in core called updateFrontmatter() which you can use here instead of this

packages/chatbot-server-mongodb-public/src/processors/formatUserMessageForGeneration.ts
```diff
@@ -0,0 +1,48 @@
+/**
+  Adds front matter to the user message text.
+ */
+import { ConversationCustomData, logger } from "mongodb-rag-core";
+
+export function formatUserMessageForGeneration(
+  userMessageText: string,
+  customData?: ConversationCustomData
+) {
+  const frontMatters: Record<string, string> = {};
```
mongodben, line 10: nit: `frontmatter` singular pls

packages/chatbot-server-mongodb-public/src/processors/formatUserMessageForGeneration.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  } catch (e) {
+    logger.warn(`Origin ${origin} malformed. Not using in front matter.`);
+  }
+
+  if (originCode === "VSCODE") {
+    frontMatters.client = "MongoDB VS Code plugin";
+  } else if (originCode === "GEMINI_CODE_ASSIST") {
+    frontMatters.client = "Gemini Code Assist";
+  }
```
mongodben, line 31: rather than hard coding these values here, to be more maintainable, can you add them to the data structure of the origin codes wherever thats defined. 

also, perhaps we should just have a client label for all origin codes, e.g. 'MongoDB Docs website', "MongoDB University", etc.

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.ts
```diff
@@ -85,7 +88,7 @@ export function makeGenerateResponseWithSearchTool({
     }
     const userMessage: UserMessage = {
       role: "user",
-      content: latestMessageText,
+      content: formatUserMessageForGeneration(latestMessageText, customData),
```
mongodben, line 91: as discussed above, i recommend validating the data that you pass as the second argument w/ zod rather than passing custom data directly. better boundries between logic + cleaner code

packages/chatbot-server-mongodb-public/src/systemPrompt.ts
```diff
@@ -31,21 +32,35 @@ const technicalKnowledge = [
 ];
 
 const importantNotes = [
-  `ALWAYS use the ${SEARCH_TOOL_NAME} tool at the start of the conversation. Zero exceptions!`,
-  `Use the ${SEARCH_TOOL_NAME} tool after every single user message.`,
+  "User messages may be accompanied by metadata explaining where the chatbot is on the site.",
```
mongodben, line 35: my prompting intuition tells me that this should be a separate section of the prompt, rather than just putting in importantNotes.

can create new section of prompt like:

```txt
<message-metadata>
User messages may be accompanied by metadata as follows:
{{list of metadata rules}}
</message-metadata>

```

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
```diff
@@ -0,0 +1,111 @@
...
+              $ne: "deleted",
+            },
+          },
+        });
+
+        let text: string;
+        let reference: Reference | undefined;
+        if (page === null) {
+          // Fall back - no page for this URL
+          text = "{fallback_to_search}";
```
mongodben, line 65: why not just handle the search inside of the tool? will be more performant + lower risk of confusing model. and in the tool call response, you can just note that it couldn't find the page, but performed the fallback to general search

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
```diff
@@ -0,0 +1,111 @@
...
+        } else if (page.body.length < 150000) {
```
mongodben, line 66: can u put the 150000 magic number as argument in the makeFetchPageTool constructor?

generally good for magic numbers to be configurable

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
```diff
@@ -0,0 +1,111 @@
...
+          query: {
+            action: {
+              $ne: "deleted",
+            },
+          },
+        });
+
+        let text: string;
+        let reference: Reference | undefined;
+        if (page === null) {
```
mongodben, line 63: for readability, i rec moving the if/else if/else logic below to a separate helper function that returns `{text: string, reference: Reference | undefined}`

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
mongodben: while, per above comment, i think you should refactor this approach, generally speaking it's an antipattern to include a magic value like this directly in the code. better for it to be a top-level defined variable that you reference here. eg:

```ts
// top of file
const fallbackText = "{fallback_to_search}";

// ...
// ..when you reference it
text = fallbackText;

```

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
mongodben: we like `const` in JS world 😄

packages/mongodb-rag-core/src/contentStore/EmbeddedContent.ts
mongodben, line 1: on the function makeMongoDbEmbeddedContentStore, pls add the new url index to the index construction

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.ts
mongodben, line 1: wow love how little code changes needed here

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.test.ts
mongodben, line 1: should there be at least 1 test for the fetchPage tool call?

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
```diff
@@ -0,0 +1,111 @@
...
+});
+
+export type MongoDbFetchPageToolArgs = z.infer<
+  typeof MongoDbFetchPageToolArgsSchema
+>;
+
+export type FetchPageToolResult = {
+  url: string;
+  text: string;
+  reference?: Reference;
```
mongodben, line 22: per below comment about the search fall back, consider doing `references: References` (plural). then u can also remove the `url` field.

mongodben (commented):
excellent start!! looking forward to seeing how this performs.

since this is merged, i cant leave a proper review, but requesting changes!

packages/chatbot-server-mongodb-public/src/tools/fetchPage.ts
hschawe: ooh that makes sense, i was not a fan of using `let` outside of the if statements but i wasn't sure about alternatives

PR #798, opened 2025-07-01
(EAI-1082) Fix missing scrubbed messages

nlarew:
Jira: (EAI-1082) Fix missing scrubbed messages

## Changes

- Runs "update trace handlers" for both streamed and non-streamed responses (previously only ran for streamed)


mongodben (approved):
lgtm (and my bad 😰)

PR #814, opened 2025-07-10
EAI-1106 Braintrust evals on fetch_page

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1106

## Changes

- Evals on generateResponseWithTools - checking if the right tool is selected, and that the correct args are passed to the tool
- Evals on conversations.eval.ts - validates quality of response generated after fetch_page is used 

## Braintrust Evaluation Results

[Conversation Results ](https://www.braintrust.dev/app/mongodb-education-ai/p/mongodb-chatbot-conversations/experiments?tg=false&search={%22filter%22:[%22id%2520%253D%2520%279bd7891f-86ca-4cbf-b300-cd075dd83ef4%27%2520OR%2520%250Aid%2520%253D%2520%2717399d9f-fbc5-4956-8f16-e6fc286d47c6%27%2520OR%2520%250Aid%2520%253D%2520%278e0bdc4f-7e4d-4c7c-a6e0-1eca971bf3c3%27%2520OR%2520%250Aid%2520%253D%2520%279a12e25b-fcfa-4b6c-b41c-c8fa26f75bd6%27%2520OR%2520%250Aid%2520%253D%2520%27e6eadb01-9f1f-4b2b-aa3e-9356bd1ce2c8%27%22]})
- contextRelevancy doesn't feel like a good metric to use here / it's only meaningful when comparing fetch_page results to each other, since they have much more context = much lower relevancy scores 
- 150k, 100k character page truncation length is fine, a response quality drop is only noticed around the 50k cutoff length
- the "always search" experiment also slightly improved Factuality & Faithfulness results

[ generateResponseWithTools Results ](https://www.braintrust.dev/app/mongodb-education-ai/p/mongodb-chatbot-generate-with-tools/experiments) 
- %MessageOrderCorrect and %ToolArgumentsCorrect are high, so the correct tool (search/fetch_page) is being used for our cases 
- I included 2 cases on whether the system would fallback to search if the context retrieved by fetch_page didn't contain info to answer the user question (even though this wasn't in-scope). They both fail to run the search tool


packages/chatbot-server-mongodb-public/evalCases/conversations.yml
```diff
@@ -1476,3 +1476,216 @@
...
+- name: should not include URLs in the response even if user asks for them (fetch_page)
+  messages:
+    - role: user
+      content: Can you give me the link to the documentation for capped collections? Use this page https://www.mongodb.com/docs/manual/core/capped-collections/
+  expectation: >
+    The ASSISTANT does NOT include any URLs in the response.
+  reject: false
+  tags:
+    - fetch_page
+- name: should not include URLs in the response even if user asks for them (fetch_page) 2
```
mongodben, line 1674: this doesn't seem like desirable behavior, even if its the current expected behavior. 

maybe worth rethinking how we handle questions like this. no action needed now. just food for thought

packages/chatbot-server-mongodb-public/src/eval/getGenerateWithToolEvalCasesFromYaml.ts
mongodben, line 1: nice. i like this.

i wonder if at some point we ought to consolidate this w/ the existing ConversationEvalCase type. did you think about this option? curious why you decided to make separate

(not requesting changes here..if we change can do in the upcoming eval improvement work)

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+  messages: GenerateResponseExpectedMessage[];
+};
+
+// Load eval cases from a YAML file.
+const evalCases: EvalCase<
+  GenerateResponseInput,
+  GenerateResponseExpected,
+  IntermediateToolResponse
+>[] = getGenerateWithToolsEvalCasesFromYamlFile(
+  require.resolve("../../evalCases/generate_response_with_tools.yml")
```
mongodben, line 56: this is a bit of an odd pattern for loading the eval cases. was there a specific reason you chose?

generally prefer using `path` module (`path.resolve(...)`). see other examples in repo for reference, and pls update this to use as well if there wasn't a clear reason for this one.

also as a general practice when loading stuff from file, prefer not to use `/` b/c if using this code on windows, i believe it wouldnt work b/c windows uses `\`. instead do `path.resolve("..", "..", "evalCases" ...)`

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+        };
+      }
+      if (
+        expectedMessage.toolCallName !== outputMessage.toolCall?.function.name
+      ) {
+        return {
+          name: "CorrectToolCall",
+          score: 0,
+          metadata: {
+            message: `Assistant returned the wrong tool call`,
```
mongodben, line 121:
```suggestion
            message: `Assistant returned the wrong tool call`,
            actualToolCall: outputMessage.toolCall?.function.name
```

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+
+/** Verify the correct args were generated for the tool call(s) */
+const ScoreToolsUsedCorrectly: EvalScorer<
+  GenerateResponseInput,
+  GenerateResponseReturnValue,
+  GenerateResponseExpected
+> = ({ output, expected }) => {
+  let totalToolArgsCorrect = 0;
+  let totalToolArgs = 0;
+  for (
```
mongodben, line 141: the logic here seems fairly duplicative of the above scorer. 

i suggest creating one helper function that extracts all the necessary data for both the scorers, and then the logic in the scorer itself can be much simpler

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+// Create dependencies
+const azureOpenAi = createAzure({
+  apiKey: OPENAI_API_KEY,
+  resourceName: process.env.OPENAI_RESOURCE_NAME,
+});
+const languageModel = azureOpenAi(OPENAI_CHAT_COMPLETION_DEPLOYMENT);
+
+// Run the eval. We recreate generateResponseWithTools each time so
+// we can pass different intermediate return values for lookup/search tools
+Eval("mongodb-chatbot-generate-w-tools", {
```
mongodben, line 204:
```suggestion
Eval("mongodb-chatbot-generate-with-tools", {
```

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+  data: evalCases,
+  experimentName: "mongodb-chatbot-generate-w-tools",
+  maxConcurrency: 10,
```
mongodben, line 207: i think you can probbably go higher on this. we have pretty high rate limits. see the default max concurrency in the `models` file for gpt-4.1 in mongodb rag core

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+        format: "md",
+        sourceName: "<This is the source name>",
+        metadata: {},
+        title: "<This is the page title>",
+        updated: new Date(),
+        action: "updated",
+      } satisfies PersistedPage;
+    };
+
+    const generateResponseWithTools = makeGenerateResponseWithTools({
```
mongodben, line 248: for these arguments, rather than defining everything here, can u instead import from `config`? then do an override for anything you need to replace, like the LLM? 

that way things don't get out of date.

packages/chatbot-server-mongodb-public/src/tracing/extractTracingData.ts
mongodben, line 1: can u summarize what changes were to this file? 

and should there be any new tests added?

packages/chatbot-server-mongodb-public/evalCases/conversations.yml
hschawe: yeah we've also had some chatbot feedback on the same point as well

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.eval.ts
```diff
@@ -0,0 +1,288 @@
...
+        findContent: mockFindContent,
+        makeReferences: makeMongoDbReferences,
+      }),
+      fetchPageTool: makeFetchPageTool({
+        loadPage,
+        findContent: mockFindContent,
+        makeReferences: makeMongoDbReferences,
+      }),
+      toolChoice: "auto",
+      maxSteps: 5,
```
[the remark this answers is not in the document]
hschawe: toolChoice and maxSteps are the same as what's used in `config.ts`, i've changed it to import these params from there instead

packages/chatbot-server-mongodb-public/src/eval/getGenerateWithToolEvalCasesFromYaml.ts
mongodben: > we could consolidate the types, i think it could get pretty bloated though... maybe we could change ConversationEvalCase to be more of a "base" and then extend it for the different cases? 

this seems like a reasonable pattern

packages/ingest-mongodb-public/src/sources/mongodb-university/MongoDbUniversityDataApiClient.ts
```diff
@@ -201,7 +201,7 @@ export function makeMongoDbUniversityDataApiClient({
     },
     async getAllVideos() {
       let offset = 0;
-      const LIMIT = 300; // Maximum allowed by the MongoDB University Data API.
+      const LIMIT = 250; // Maximum allowed by the MongoDB University Data API.
```
mongodben, line 204: revert pls

PR #817, opened 2025-07-14
additional secrets required for dependency (benchmarks) build

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-1119

## Changes

- script depends on benchmarks and benchmarks package needs secrets to build
-

## Notes

-


mongodben (approved):
pls make sure the BT api key that you're using is separate from the one you use on your laptop

PR #820, opened 2025-07-16
(EAI-1138) Ingest ILT schedule

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1138

## Changes

- Ingest instructor-led training schedule page 

## Notes

- to ingest ILT course detail PDFs in a separate PR


mongodben (approved):
this content structure is pretty different from what we normally ingest. wonder did u see well the ingestion works on it?

frankly i dont think this content merits large work itself, but could be a good opportunity for us to improve the web scraping more generally

hschawe:
@mongodben ingesting this page works fine for the most part (the markdown tables are pulled, but some of the links are not) but the table contents may not play nice with he bot because the table cells contain pipe symbols. either way good point

PR #821, opened 2025-07-16
EAI-1168 Remove query string, hash fragment in normalizeUrl

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1168

## Changes

- Option to remove query string & hash fragment in normalizeUrl

## Notes

- I made these configurable because there could be future cases where we don't want to remove them (e.g. before, in ingest, we had 4 unique pages for the same URL, only difference was the URL hash fragments)

### DIFF
--- a/packages/ingest-mongodb-public/src/sources/mongodbDotCom/webSources.ts
+++ b/packages/ingest-mongodb-public/src/sources/mongodbDotCom/webSources.ts
@@ -26,6 +26,12 @@ export type InitialWebSource = {
    Optional additional metadata determined by the web source.
    */
   staticMetadata?: Record<string, string | string[]>;
+
+  /**
+    URL normalization options for this source.
+    */
+  removeHash?: boolean;
+  removeQueryString?: boolean;
 };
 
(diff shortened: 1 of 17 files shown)


yakubova92:
> logically this looks correct, but i think there's some stuff for university where we want either the QS or fragment. think @yakubova92 should have context

Skills specifically comes to mind. Skills pages have urls like this one for example: https://learn.mongodb.com/skills?openTab=gen+ai, where if the query param is not present, the content will not get picked up by the scraper

packages/ingest-mongodb-public/src/sources/mongodbDotCom/webSources.ts
```diff
@@ -26,6 +26,12 @@ export type InitialWebSource = {
    Optional additional metadata determined by the web source.
    */
   staticMetadata?: Record<string, string | string[]>;
+
+  /**
+    URL normalization options for this source.
+    */
+  removeHash?: boolean;
+  removeQueryString?: boolean;
```
hschawe, line 34: Made the normalization options definable here in websources

PR #828, opened 2025-07-18
Implement generate new response

ASteinheiser:
Jira: https://jira.mongodb.org/browse/EAI-125

## Changes

- Configure the public chat server to use the new responses streaming API

## Notes

-

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/processors/generateResponseWithSearchTool.ts
+++ b/packages/chatbot-server-mongodb-public/src/processors/generateResponseWithSearchTool.ts
@@ -95,7 +95,13 @@ export const addMessageToConversationStream: GenerateResponseWithSearchToolParam
 export const responsesApiStream: GenerateResponseWithSearchToolParams["stream"] =
   {
     onLlmNotWorking({ dataStreamer, notWorkingMessage }) {
-      // only stream "done" here since it's one message
+      dataStreamer?.streamResponses({
+        type: "response.output_text.delta",
+        delta: notWorkingMessage,
+        content_index: 0,
+        output_index: 0,
+        item_id: "",
+      } satisfies ResponseStreamOutputTextDelta);
       dataStreamer?.streamResponses({
         type: "response.output_text.done",
         text: notWorkingMessage,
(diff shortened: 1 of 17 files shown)


packages/chatbot-server-mongodb-public/src/processors/generateResponseWithSearchTool.ts
```diff
@@ -88,20 +89,56 @@ export const addMessageToConversationStream: GenerateResponseWithSearchToolParam
...
-      throw new Error("not yet implemented");
+    onTextDelta({ dataStreamer, delta }) {
+      // only stream delta here, allow "done" to be streamed elsewhere
+      dataStreamer?.streamResponses({
+        type: "response.output_text.delta",
+        delta,
+        content_index: 0,
+        output_index: 0,
+        item_id: "",
+      } as OpenAI.Responses.ResponseTextDeltaEvent);
```
mongodben, line 141: no `as`!! satisifes only pls

mongodben: if still weirdness, see if can do pattern like:

```ts
({
  //...
} satisfies SubType) as OpenAI.Repsonse.ResponseStreamEvent
```

i feel pretty strongly that we should have some level of typeguarding here, even if we also need to use type coercion

ASteinheiser: yeah agreed! and that makes sense. I'll give that satisfies/as approach a shot!

ASteinheiser: I think I found a good solution for this that uses `satisfies` and it's too cumbersome!

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithSearchTool.ts
```diff
@@ -88,20 +92,65 @@ export const addMessageToConversationStream: GenerateResponseWithSearchToolParam
     },
   };
 
-// TODO: implement this
 export const responsesApiStream: GenerateResponseWithSearchToolParams["stream"] =
   {
-    onLlmNotWorking() {
-      throw new Error("not yet implemented");
+    onLlmNotWorking({ dataStreamer, notWorkingMessage }) {
+      // only stream "done" here since it's one message
```
mongodben, line 98: should there not also be a delta w/ the content before `done`?

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithSearchTool.ts
```diff
@@ -88,20 +92,65 @@ export const addMessageToConversationStream: GenerateResponseWithSearchToolParam
...
-    onTextDelta() {
-      throw new Error("not yet implemented");
+    onTextDone({ dataStreamer, text }) {
+      dataStreamer?.streamResponses({
+        type: "response.output_text.done",
+        text,
+        content_index: 0,
+        output_index: 0,
+        item_id: "",
+      } satisfies ResponseStreamOutputTextDone);
```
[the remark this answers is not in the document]
mongodben: i think output text is just for annotations https://platform.openai.com/docs/api-reference/responses-streaming/response/output_text

mongodben:
> I think I just need to figure out how to properly test the chat-server-public now

can't you just use supertest `request()` module?

PR #833, opened 2025-07-21
(EAI-1173) Content sources endpoint

mmeigs:
Jira: https://jira.mongodb.org/browse/EAI-1173

## Changes

- Adds new GET `/content/sources` endpoint to list all data sources possible
- Adds route handler and tests for route handler
- Adds documentation for new route in openapi yaml

### DIFF
--- a/packages/mongodb-chatbot-server/src/routes/content/listDataSources.test.ts
+++ b/packages/mongodb-chatbot-server/src/routes/content/listDataSources.test.ts
@@ -1,4 +1,7 @@
-import type { MongoDbEmbeddedContentStore, DataSourceMetadata } from "mongodb-rag-core";
+import type {
+  MongoDbEmbeddedContentStore,
+  DataSourceMetadata,
+} from "mongodb-rag-core";
 import { createRequest, createResponse } from "node-mocks-http";
 import { makeListDataSourcesRoute } from "./listDataSources";
 
@@ -10,8 +13,19 @@ function makeMockEmbeddedContentStore(dataSources: DataSourceMetadata[]) {
 
 describe("makeListDataSourcesRoute", () => {
   const mockDataSources: DataSourceMetadata[] = [
-    { id: "source1", version: ["current", "v6.0"], type: "docs" },
-    { id: "source2", version: ["v2.11"], type: "university-content" },
+    {
+      id: "source1",
+      version: [
+        { label: "current", isCurrent: true },
+        { label: "v6.0", isCurrent: false },
+      ],
+      type: "docs",
+    },
+    {
+      id: "source2",
+      version: [{ label: "v2.11", isCurrent: false }],
+      type: "university-content",
+    },
   ];
 
   it("should return data sources for a valid request", async () => {
--- a/packages/mongodb-chatbot-server/src/routes/content/listDataSources.test.ts
+++ b/packages/mongodb-chatbot-server/src/routes/content/listDataSources.test.ts
@@ -1,0 +1,72 @@
+import type { MongoDbEmbeddedContentStore, DataSourceMetadata } from "mongodb-rag-core";
+import { createRequest, createResponse } from "node-mocks-http";
+import { makeListDataSourcesRoute } from "./listDataSources";
+
+function makeMockEmbeddedContentStore(dataSources: DataSourceMetadata[]) {
+  return {
+    listDataSources: jest.fn().mockResolvedValue(dataSources),
+  } as unknown as MongoDbEmbeddedContentStore;
+}
+
+describe("makeListDataSourcesRoute", () => {
+  const mockDataSources: DataSourceMetadata[] = [
+    { id: "source1", version: ["current", "v6.0"], type: "docs" },
+    { id: "source2", version: ["v2.11"], type: "university-content" },
+  ];
+
+  it("should return data sources for a valid request", async () => {
+    const embeddedContentStore = makeMockEmbeddedContentStore(mockDataSources);
+    const handler = makeListDataSourcesRoute({ embeddedContentStore });
+
+    const req = createRequest({
+      headers: { "req-id": "test-req-id" },
+    });
+    const res = createResponse();
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    await handler(req, res as any);
+
+    const data = res._getJSONData();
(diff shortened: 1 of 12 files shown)


packages/mongodb-rag-core/src/contentStore/EmbeddedContent.ts
```diff
@@ -101,6 +101,15 @@ export type QueryFilters = {
   sourceType?: Page["sourceType"] | string[];
 };
 
+/** 
+  Metadata of data source
+*/
+export type DataSourceMetadata = {
+  id: string;
+  version?: string[];
```
mongodben, line 109:
```suggestion
  version?: {label: string, isCurrent: boolean }[];
```

pls track the current version in here as well. something like the above typing

mongodben (changes requested):
generally looks good. see note about adding the isCurrent info

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -274,6 +275,59 @@ export function makeMongoDbEmbeddedContentStore({
...
+    async listDataSources(): Promise<DataSourceMetadata[]> {
+      return await embeddedContentCollection
+        .aggregate<DataSourceMetadata>([
+          {
+            $group: {
+              _id: "$sourceName",
+              version: {
+                $addToSet: {
+                  label: "$metadata.version.label",
+                  isCurrent: "$metadata.version.isCurrent",
```
mongodben, line 287: ❤️

packages/mongodb-chatbot-server/src/routes/content/listDataSources.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const handler = makeListDataSourcesRoute({ embeddedContentStore });
+
+    const req = createRequest({
+      headers: { "req-id": "test-req-id" },
+    });
+    const res = createResponse();
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    await expect(handler(req, res as any)).rejects.toMatchObject({
+      message: "Unable to list data sources",
```
mongodben, line 63: see note on listDataSources.ts about this

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -274,6 +275,59 @@ export function makeMongoDbEmbeddedContentStore({
       }
     },
 
+    async listDataSources(): Promise<DataSourceMetadata[]> {
```
mongodben, line 278: pls add some basic tests for this method to `MongoDbEmbeddedContentStore.test.ts`. can use `mongodb-memory-server` for these tests to keep simple.

mongodben (changes requested):
looks great. few small things

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -274,6 +275,59 @@ export function makeMongoDbEmbeddedContentStore({
...
+      return await embeddedContentCollection
+        .aggregate<DataSourceMetadata>([
```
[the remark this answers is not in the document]
mongodben: we can re-approach how to efficiently do this operation in a meeting. @nlarew will probably be better at thinking this through than me

mongodben (commented):
lets get a review from @nlarew on the performance stuff

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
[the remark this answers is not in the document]
mongodben: as discussed in slack, there might be a compound index solution. or possibly we could do some caching on the server of precomputed results, and just refresh it every so often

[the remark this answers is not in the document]
nlarew: I think the compound index solution is likely sufficient. Looks like we already have the proper index defined on dev.

Note - `sourceType` will be 1:1 with `sourceName` so we can use `$first` and simplify the pipeline just a bit.

WIth the compound index this pipeline takes ~250ms which I think is fine personally. If we're really concerned about performance still then we can consider a server cache with, say, 1 hour retention.

```js
db.runCommand({
  explain: {
    aggregate: "embedded_content",
    pipeline: [
      {
        $group: {
          _id: "$sourceName",
          version: {
            $addToSet: {
              label: "$metadata.version.label",
              isCurrent: "$metadata.version.isCurrent"
            }
          },
          sourceType: { $first: "$sourceType" }
        }
      },
      {
        $project: {
          _id: 0,
          id: "$_id",
          version: {
            $map: {
              input: {
                $filter: {
                  input: "$version",
                  as: "v",
                  cond: { $ne: ["$$v.label", null] }
                }
              },
              as: "v",
              in: {
                label: "$$v.label",
                isCurrent: { $ifNull: ["$$v.isCurrent", false] }
              }
            }
          },
          type: "$sourceType"
        }
      }
    ],
    cursor: {}
  },
  verbosity: "executionStats"
})
```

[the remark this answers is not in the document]
nlarew: Oh yeah on further investigation the `$addToSet` basically prevents us from using an IXSCAN. If you that piece then it does a DISTINCT_SCAN which is even more performant than in IXSCAN.

I'm sure there's a well thought out correctness guarantee behind this but that's pretty annoying.

packages/mongodb-rag-core/src/contentStore/MongoDbEmbeddedContentStore.ts
```diff
@@ -274,6 +353,36 @@ export function makeMongoDbEmbeddedContentStore({
       }
     },
 
+    async listDataSources(): Promise<DataSourceMetadata[]> {
+      const now = Date.now();
+      const maxAge = 1000 * 60 * 60 * 24 * 7; // 7 days
```
mongodben, line 358: for this and the other magic numbers, pls define as constants at the top of the file

PR #838, opened 2025-07-23
(EAI-1119) add category, clearer logging, cron runs on dev db too

yakubova92:
Jira: https://jira.mongodb.org/browse/EAI-EAI-1119

## Changes

- category pulled from related llm_cases doc and saved to llm_answers doc
- improved logging
- `process-citations-for-metrics-dev` in staging cron will run this script against the dev db so that we have up to date info on dev, since we're always displaying yesterday's stats

## Notes

-

### DIFF
--- a/packages/scripts/src/profound/getAndProcessAnswers.ts
+++ b/packages/scripts/src/profound/getAndProcessAnswers.ts
@@ -193,7 +193,7 @@ export const main = async (startDateArg?: string, endDateArg?: string) => {
   };
   const referenceAlignmentFn = makeReferenceAlignment(openAiClient, config);
   const answerRecords: any[] = [];
-  const answerRecordsWithNoAssociatedCase = new Set()
+  const promptsWithNoAssociatedCase = new Set()
   const { results, errors } = await PromisePool.for(answers)
     .withConcurrency(model.maxConcurrency ?? 5)
     .process(async (currentAnswer) => {
@@ -209,7 +209,7 @@ export const main = async (startDateArg?: string, endDateArg?: string) => {
       const currentPromptId = currentAnswer.prompt_id;
       const currentCase = casesByPromptMap[currentPromptId];
       if (!currentCase) {
-        answerRecordsWithNoAssociatedCase.add(`${currentPromptId} - ${currentPrompt}`);
+        promptsWithNoAssociatedCase.add(`${currentPromptId} - ${currentPrompt}`);
       }
 
       // calculate reference alignment score
@@ -295,8 +295,8 @@ export const main = async (startDateArg?: string, endDateArg?: string) => {
       return answerEngineRecord;
     });
 
-  console.log(`No cases found for ${answerRecordsWithNoAssociatedCase.size} prompts:`)
-  answerRecordsWithNoAssociatedCase.forEach((promptInfo: any) => {
+  console.log(`Found ${promptsWithNoAssociatedCase.size} prompts with no associated case:`)
+  promptsWithNoAssociatedCase.forEach((promptInfo: any) => {
     console.log(` - ${promptInfo}`);
   });
   // update the llm_answers collection


packages/scripts/environments/production.yml
```diff
@@ -57,7 +57,7 @@ cronJobs:
     backoffLimit: 2
 
   - name: process-citations-for-metrics
-    schedule: "0 3 * * *" # every day at 3 AM UTC
+    schedule: "1 0 * * *" # every day at 00:01 (12:01 AM) UTC
```
nlarew, line 60: Do we know roughly when Profound runs their script each day? Or does that not matter because this is results for the previous day?

packages/scripts/environments/staging.yml
```diff
@@ -86,6 +86,27 @@ cronJobs:
         memory: 5Gi
     backoffLimit: 2
 
+  - name: process-citations-for-metrics-dev
+    schedule: "30 0 * * *" # every day at 00:30 (12:30 AM) UTC
```
nlarew, line 90: Any reason for using a different time than in prod? Probably doesn't really matter but might be nice to be consistent.

packages/scripts/environments/staging.yml
```diff
@@ -86,6 +86,27 @@ cronJobs:
...
+    command: ["npm", "run", "scripts:getLLMAnswers"]
+    env:
+      MONGODB_DATABASE_NAME: docs-chatbot-dev
+      PROFOUND_API_URL: https://api.tryprofound.com/v1
+      BRAINTRUST_ENDPOINT: https://api.braintrust.dev/v1/proxy
+    envSecrets:
+      MONGODB_CONNECTION_URI: docs-chatbot-scripts-admin-dev
```
nlarew, line 97: Is `docs-chatbot-scripts-admin-dev` a new secret in k8s? Just confirming that's intentional + making sure you added the secrets to k8s.

packages/scripts/src/profound/getAndProcessAnswers.ts
```diff
@@ -283,12 +288,17 @@ export const main = async (startDateArg?: string, endDateArg?: string) => {
...
         profoundRunId: currentAnswer.run_id,
-        dataset: currentCase ? getDataset(currentCase.tags, datasetsByTagMap) : null
+        dataset: currentCase ? getDataset(currentCase.tags, datasetsByTagMap) : null,
+        category: currentCase ? currentCase.metadata.category : null
       };
       answerRecords.push(answerEngineRecord);
       return answerEngineRecord;
     });
 
+  console.log(`No cases found for ${answerRecordsWithNoAssociatedCase.size} prompts:`)
```
nlarew, line 298:
```suggestion
  console.log(`Found ${answerRecordsWithNoAssociatedCase.size} answer engine results with no corresponding case:`)
```

nlarew (approved):

packages/scripts/environments/production.yml
yakubova92: Based on the timestamps we have, it looks like Profound is running scripts all through the day. We query Profound for the previous day, so all records should be in. We're also using UTC everywhere bc the timestamps on Profound's records are UTC

packages/scripts/environments/staging.yml
yakubova92: it is an existing secret, where MONGODB_CONNECTION_URI points to the dev db

packages/scripts/src/profound/getAndProcessAnswers.ts
yakubova92: changed this to `promptsWithNoAssociatedCase`, as these aren't actually answer records. Im taking the prompt text and the prompt id from any answer records that we can't find an associated course for and adding to a Set. The prompt is run against multiple platforms, so we end up with a really long list of answer engine response, but really only a few prompts.

PR #839, opened 2025-07-23
Responses api

ASteinheiser:
Jira: https://jira.mongodb.org/browse/EAI-945

## Changes

- add new responses API, based off the [openai spec](https://platform.openai.com/docs/api-reference/responses)

## Notes

- this is a basic version and many inputs are not respected. You should still receive answers, but expect some settings to be no-ops for now. Updates coming soon

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/responsesApi.test.ts
+++ b/packages/chatbot-server-mongodb-public/src/responsesApi.test.ts
@@ -15,7 +15,6 @@ const API_PREFIX = "/api/v1";
 
 // Response event types
 const DELTA_EVENT = "response.output_text.delta";
-const ANNOTATION_EVENT = "response.output_text.annotation.added";
 const DONE_EVENT = "response.output_text.done";
 
 describe("Responses API with OpenAI Client", () => {
@@ -284,9 +283,8 @@ describe("Responses API with OpenAI Client", () => {
       const doneCount = events.filter((t) => t === DONE_EVENT).length;
 
       // 3 lifecycle events are validated in the expectValidResponses function above
-      // 3 lifecycle events + 21 delta events + 1 done event = 25 events
-      expect(responses.length).toBe(25);
-      expect(deltaCount).toBe(21);
+      expect(responses.length).toBeGreaterThanOrEqual(5);
+      expect(deltaCount).toBeGreaterThanOrEqual(1);
       expect(doneCount).toBe(1);
     });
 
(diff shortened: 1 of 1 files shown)


packages/chatbot-server-mongodb-public/package.json
```diff
@@ -29,6 +29,7 @@
   "dependencies": {
     "@segment/analytics-node": "^2.2.1",
     "@slack/web-api": "^7.8.0",
+    "@types/supertest": "^6.0.3",
```
mongodben, line 32: can you move this and `supertest` to dev dependencies?

packages/mongodb-chatbot-server/src/processors/makeVerifiedAnswerGenerateResponse.test.ts
```diff
@@ -24,6 +27,29 @@ describe("makeVerifiedAnswerGenerateResponse", () => {
...
+    });
+    dataStreamer.streamData({
+      type: "delta",
+      data: verifiedAnswer.answer,
+    });
+    dataStreamer.streamData({
+      type: "references",
+      data: verifiedAnswer.references,
+    });
+  };
```
ASteinheiser, line 51: @mongodben  in reference to that case I mentioned about "not receiving a finished event when references are returned", do you think that's related to this? Should this stream a finished event?

packages/mongodb-chatbot-server/src/processors/makeVerifiedAnswerGenerateResponse.ts
```diff
@@ -17,8 +21,40 @@ export interface MakeVerifiedAnswerGenerateResponseParams {
...
+      });
+      dataStreamer.streamData({
+        type: "delta",
+        data: verifiedAnswer.answer,
+      });
+      dataStreamer.streamData({
+        type: "references",
+        data: verifiedAnswer.references,
+      });
+    },
```
ASteinheiser, line 55: same here?

packages/mongodb-chatbot-server/src/processors/makeVerifiedAnswerGenerateResponse.ts
mongodben, line 1: do we need to add the responses version of the streaming for this?

packages/mongodb-chatbot-server/src/routes/responses/responsesRouter.ts
```diff
@@ -0,0 +1,63 @@
...
+    Slow down the response to the responsesRouter after certain number
+    of requests in the time window.
+   */
+  const globalSlowDown = makeSlowDown(rateLimitConfig?.routerSlowDownConfig);
+  responsesRouter.use(globalSlowDown);
+
+  // Create Response API
+  responsesRouter.post("/", makeCreateResponseRoute(createResponse));
+
+  return responsesRouter;
```
ASteinheiser, line 62: @mongodben just double checking here, do we need any other middlewares applied at this level? I don't think any others are being explicitly applied on top of the responses route besides these

mongodben (changes requested):
1 thing i noticed is that we likely will need to do similar streaming stuff that we do on the main generate response function to verified answer. see comment.

also note about using dev dependencies

packages/mongodb-chatbot-server/src/processors/makeVerifiedAnswerGenerateResponse.ts
mongodben: yes pls!

packages/mongodb-chatbot-server/src/routes/responses/responsesRouter.ts
ASteinheiser: sweet I think we're good then!

packages/mongodb-chatbot-server/src/processors/makeVerifiedAnswerGenerateResponse.ts
ASteinheiser: this actually is the solution for my two comments below!

PR #840, opened 2025-07-24
(EAI-1175) Add slot to messages below message rating

nlarew:
Jira: (EAI-1175) Add slot to messages below message rating

## Changes

- Adds `View.messageBottomContent` prop to slot in arbitrary react nodes at the bottom of each message
- Will be used by forum team for links out to 3PF


mongodben (approved):
rubber-stamp

PR #842, opened 2025-07-24
CORS for localhost in non-prod

mongodben:
Jira: n/a

## Changes

- Add support for reqs from localhost for non-production environments

## Notes

-


ASteinheiser (approved):
nice!

PR #844, opened 2025-07-24
Implement case analysis script

cbush:
This adds the `assessCases` script that analyzes the prompt/expected answer pairs using two methodologies:

1. [Answer relevance](https://docs.ragas.io/en/v0.1.21/concepts/metrics/answer_relevance.html): given expected answer, generate N prompts that could elicit that answer, then compare their embeddings with the embedding of the original given prompt
2. LLM as judge: use LLM to score the prompt/answer pair on a variety of metrics and to generate recommendations for improvement.


packages/scripts/src/SimpleTextGenerator.ts
```diff
@@ -0,0 +1,49 @@
+import { OpenAI } from "mongodb-rag-core/openai";
```
mongodben, line 1: i recommend using the AI SDK's function `generateText()` for this. wont need to make own constructor. can import from `mongodb-rag-core/aiSdk`. there's lots of examples throughout the project for usage

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+  RelevanceMetrics,
+  ScoredPromptAndEmbeddings,
+} from "./Case";
+import { cosineSimilarity } from "mongodb-rag-core/aiSdk";
+
+/**
+  Given the expected answer, generate a number of possible prompts that could
+  elicit that expected answer.
+ */
+export const generatePromptsFromExpectedAnswer = async ({
```
mongodben, line 19: would be good to have lite evals for this.

not sure what metric you'd use to grade outputs, but at least good to have a few test cases that you can run this over for if you change the prompt.

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+  generate,
+}: {
+  prompt: string;
+  expected: string;
+  generate: SimpleTextGenerator;
+}): Promise<LlmAsJudgment | undefined> => {
+  const shortName = makeShortName(prompt);
+  console.log(`Rating '${shortName}' with LLM...`);
+
+  const [response] = await generate({
```
mongodben, line 185: rather than raw text gen and telling it to do json, models support structured output generation, which'd be perfect for this use case.

i recommend using the AI SDK [generateObject()](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-object) function. takes schema from zod type and handles output parsing.

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+- assumption: how much domain-specific knowledge is required to effectively answer the prompt.
+- impact: the business impact/relevance of the question and answer. Good examples: competitor questions, technical questions. Bad exaples: when was MongoDB founded?
+- guidance (string, optional): TERSELY and DIRECTLY detail the issue; suggest how to improve. Only include this if the above scores are low.
+
+Now evaluate this pair, returning only the JSON object:
+
+PROMPT: ${prompt}
+---
+EXPECTED ANSWER: ${expected}
+`,
```
mongodben, line 203:
```suggestion
`.trim(),
```

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+  generate: SimpleTextGenerator;
+}): Promise<LlmAsJudgment | undefined> => {
+  const shortName = makeShortName(prompt);
+  console.log(`Rating '${shortName}' with LLM...`);
+
+  const [response] = await generate({
+    prompt: `
+Evaluate the quality of the following prompt-expected answer pair across
+multiple dimensions. Return your evaluation as a JSON object with numeric percentage scores
+from 0 (poor) to 1 (excellent) up to 3 decimal places. Return only a JSON object (NOT IN MARKDOWN) with the following keys:
```
mongodben, line 189: >  Return your evaluation as a JSON object with numeric percentage scores
from 0 (poor) to 1 (excellent) up to 3 decimal places.

having LLM generate number 0-1 seems like a suboptimal prompting strategy. how to differentiate between .5 vs .543?? instead i recommend using likert scale scores, 1-5. provide a brief overview of what each score means in the prompt

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+
+- reasonableness: how reasonable it would be to expect an LLM to produce the given response from the given prompt.
+- clarity: how well formulated and clear the prompt is.
+- fit: how well the expected answer actually matches the prompt.
+- assumption: how much domain-specific knowledge is required to effectively answer the prompt.
+- impact: the business impact/relevance of the question and answer. Good examples: competitor questions, technical questions. Bad exaples: when was MongoDB founded?
+- guidance (string, optional): TERSELY and DIRECTLY detail the issue; suggest how to improve. Only include this if the above scores are low.
```
mongodben, line 196:
```suggestion
- answer_reasonableness: how reasonable it would be to expect an LLM to produce the given response from the given prompt.
- prompt_clarity: how well formulated and clear the prompt is.
- answer_fit: how well the expected answer actually matches the prompt.
- prompt_knowledge_assumption: how much domain-specific knowledge is required to effectively answer the prompt.
- business_impact: the business impact/relevance of the question and answer. Good examples: competitor questions, technical questions. Bad exaples: when was MongoDB founded?
- guidance (string, optional): TERSELY and DIRECTLY detail the issue; suggest how to improve. Only include this if the above scores are low.
```

packages/scripts/src/main/assessCasesMain.ts
```diff
@@ -0,0 +1,99 @@
...
+    OPENAI_API_VERSION: "",
+    OPENAI_RETRIEVAL_EMBEDDING_DEPLOYMENT: "",
+    CASE_COLLECTION_NAME: "",
+  });
+
+  const openAiClient = new AzureOpenAI({
+    apiKey: OPENAI_API_KEY,
+    endpoint: OPENAI_ENDPOINT,
+    apiVersion: OPENAI_API_VERSION,
+  });
```
mongodben, line 32: where is this script to be run? if just from laptop, i recommend using Braintrust proxy. it contains more model options and has caching to make reruns much faster and cheaper.

example usage here: https://github.com/mongodb/chatbot/blob/main/packages/benchmarks/src/discovery/config.ts#L34-L40

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  console.log(`Average score: ${JSON.stringify(averages)}`);
+  return {
+    prompt_embeddings: promptEmbeddings,
+    generated_prompts: scoredVariants,
+    averages,
+  };
+};
+
+export const rateWithLlm = async ({
```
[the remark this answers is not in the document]
mongodben: also, what model are you planning on using for it?

i recommend using a beefy reasoning model like o3 or gemini-2.5/claude 4 sonnet with high thinking budget.

you may also want to include a "thoughts" field in the json output, and instruct the model to provide reasoning for its scores. this lets you see what the model is 'thinking', and can help you debug the output more easily.

packages/scripts/src/assessCases.ts
```diff
@@ -0,0 +1,228 @@
...
+${JSON.stringify(judgment, undefined, 2)}`);
+    return judgment;
+  } catch (e) {
+    console.error(
+      `Failed to parse response "${response}" into LlmAsJudgment: ${
+        (e as Error)?.message
+      }`
+    );
+    return undefined;
+  }
```
mongodben, line 220: note: using vercel AI SDK generateObject() will take care of this boilerplate

packages/scripts/src/Case.ts
```diff
@@ -0,0 +1,76 @@
...
+  .object({
+    reasonableness: z.number(),
+    clarity: z.number(),
+    specificity: z.number(),
+    fit: z.number(),
+    assumption: z.number(),
+    impact: z.number(),
+    guidance: z.string(),
+  })
+  .partial();
```
mongodben, line 46: i recommend colocating this with the prompt, since if you're using structured outputs (as recommended below), this will become part of the prompt. you can also include more strict typing like .min() and .max(), plus .describe() for the model to see

packages/scripts/src/Case.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export type RelevanceMetrics = z.infer<typeof RelevanceMetrics>;
+
+export const ScoredPromptAndEmbeddings = PromptAndEmbeddings.and(
+  z.object({
+    relevance:
+      // embedding model name -> score
+      z.record(z.string(), RelevanceMetrics),
+  })
+);
```
mongodben, line 30: i dont think these zod types are actually used. instead can you just use typescript types/interfaces?

packages/scripts/src/Case.ts
```diff
@@ -0,0 +1,76 @@
...
+  embeddings: Embeddings,
+});
+
+export type PromptAndEmbeddings = z.infer<typeof PromptAndEmbeddings>;
+
+export const RelevanceMetrics = z.object({
+  // normalized square magnitude difference (lower = closer = better)
+  norm_sq_mag_diff: z.number(),
+  // cosine similarity (are vectors pointing the same way?) [-1, 1]
+  cos_similarity: z.number(),
```
mongodben, line 19: can you include in docs comments on here/somewhere what these metrics actually measure? is it how specific a prompt is to its answer? what's the difference between the 2 metrics as well?

packages/scripts/src/main/assessCasesMain.ts
nlarew: Ideally we'll run directly from a new page on the Mercury dashboard

PR #846, opened 2025-07-24
update CORS error handling

ASteinheiser:

## Changes

- update CORS error handling

### DIFF
--- a/packages/mongodb-chatbot-server/src/app.ts
+++ b/packages/mongodb-chatbot-server/src/app.ts
@@ -1,23 +1,22 @@
-import express, {
-  Express,
-  ErrorRequestHandler,
-  RequestHandler,
-  NextFunction,
-  Request as ExpressRequest,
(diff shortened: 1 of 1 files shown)


packages/mongodb-chatbot-server/src/app.ts
```diff
@@ -68,7 +67,9 @@ export interface AppConfig {
 */
 export const errorHandler: ErrorRequestHandler = (err, req, res, _next) => {
   const reqId = getRequestId(req);
-  const httpStatus = err.status || 500;
+  const isCorsError = err.message.includes("CORS");
```
nlarew, line 70: This works though would prefer if we could keep this handler decoupled from specific error types/statuses

nlarew (approved):
LGTM for a short term fix - if it's not too hard to make the error handler generic then let's go for it otherwise ship it

PR #847, opened 2025-07-25
more robust CORS error handling

ASteinheiser:
Jira: https://jira.mongodb.org/browse/EAI-1191

## Changes

- more robust CORS error handling
- always return 403 when CORS throws an error

### DIFF
--- a/packages/mongodb-chatbot-server/src/app.ts
+++ b/packages/mongodb-chatbot-server/src/app.ts
@@ -62,7 +62,7 @@ export interface AppConfig {
   expressAppConfig?: (app: Express) => Promise<void>;
 }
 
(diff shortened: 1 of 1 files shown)


packages/mongodb-chatbot-server/src/app.ts
```diff
@@ -62,14 +62,20 @@ export interface AppConfig {
   expressAppConfig?: (app: Express) => Promise<void>;
 }
 
+const corsHandler =
```
nlarew, line 65:
```suggestion
const makeCorsHandler =
```

nlarew (approved):
LGTM - let's use the `makeFoo` pattern though

PR #848, opened 2025-07-25
handle cleaning customData and metadata based on `store` setting

ASteinheiser:
Jira: https://jira.mongodb.org/browse/EAI-1158

## Changes

- handle cleaning customData and metadata based on `store` setting
- update tests

## Notes

- applies to both conversations and messages
- after some discussion, @mongodben and I agreed on going with an allowlist approach - meaning we will clear all fields unless explicitly allowed to persist regardless of `store` settings


mongodben (approved):
can you also update the server config to do the allowlist of the relevant metadata fields? pls allow list the stuff currently on the conversations collection that's not LLM/human generated

mongodben (changes requested):
sry meant to request change in previous review, not approve

packages/mongodb-chatbot-server/src/routes/responses/createResponse.ts
```diff
@@ -506,22 +511,42 @@ const formatMessage = (
...
+    metadata: formattedMetadata,
+    customData: formattedCustomData,
   };
 };
 
+/** These metadata keys will persist in conversations and messages even if store is false.
+ * Otherwise, keys will have their values set to an empty string if store is false. */
+export const ALWAYS_ALLOWED_METADATA_KEYS = ["ip", "origin", "userAgent"];
+
+interface FormatMetadataParams {
```
mongodben, line 534: this isn't passed to the constructor function, is it?

ASteinheiser: done!

ASteinheiser:
I believe only "ip", "origin", "userAgent" are stored and not user/llm generated. Any other fields missing? I checked on entries that I believe come from the new responses API. May need to update this list if we attach more metadata/custom data

packages/mongodb-chatbot-server/src/routes/responses/createResponse.test.ts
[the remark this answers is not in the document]
ASteinheiser: I made the existing tests more clear and added another to test when no keys are specified. Lmk if I missed any cases!

mongodben (approved):
nice, looks good

PR #859, opened 2025-07-30
ensure item ids are set for response events

ASteinheiser:
Jira: https://jira.mongodb.org/browse/EAI-1205

## Changes

- ensure item ids are set for response events


nlarew (commented):
Code looks fine - do we have test files that can assert these exist?

PR #860, opened 2025-07-31
Strong typing on atlas search filters

mongodben:
Jira: n/a

## Changes

- stronger typing for search filters on embedded content
- remove unused atlas search boosters

## Notes

-


mmeigs (approved):
LGTM. I don't know much about the booster, but if it's good to go, then all is well

PR #865, opened 2025-08-06
(EAI-1230): Create MDB collection and indexes for Atlas Search benchmark

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-1230

## Changes

- Script to load dataset from HF and push to MongoDB Atlas
- Add Atlas Search indexes to collection 
- Add README to datasets package explaining 

## Notes

- Did this code in Python b/c there's not a clean way to work with HF datasets from Node.js (which is quite bizarre and annoying)


packages/datasets/mongodb_datasets/config.py
```diff
@@ -0,0 +1,17 @@
+"""Configuration utilities for mongodb_datasets package."""
+
+from pathlib import Path
+from dotenv import load_dotenv
+
+# Find the project root .env file by traversing up from this file
+# Structure: packages/datasets/mongodb_datasets/config.py -> ../../../.env
+PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
+ENV_PATH = PROJECT_ROOT / ".env"
```
nlarew, line 9: I assume this works but it's a bit confusing to me:

- The example (i.e. three levels of `..` from `config.py` would resolve to `packages/.env` which doesn't exist
- The actual code calls `parent` five times which doesn't match the example and seems to me like it would be the directory about the root `chatbot` repo.

nlarew (approved):

PR #868, opened 2025-08-07
EAI-1216 Robust evals for custom instructions/tools

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1216

## Changes

- More robust evals for features added in the Responses API changes. Cases for custom system prompt, custom tools, multi-turn conversation context handling
- Centralized evals so metrics are consistent across different types of cases & it'll be easier to run them in the ci suite (upcoming work)

## Notes
- Going to handle adding guardrail cases / guardrail prompt changes in [EAI-1226](https://jira.mongodb.org/browse/EAI-1226)
- [Braintrust run](https://www.braintrust.dev/app/mongodb-education-ai/p/mongodb-chatbot-conversations/experiments/mongodb-chatbot-latest-d8ccd35e?c=mongodb-chatbot-latest-2228a2a2)


packages/chatbot-server-mongodb-public/src/eval/ConversationEval.ts
```diff
@@ -64,6 +77,7 @@ type ConversationEvalScorer = EvalScorer<
 const RetrievedContext: ConversationEvalScorer = async (args) => {
   const name = "RetrievedContext";
   if (!args.output.context) {
+    // Might want to not run this metric if custom tools are given, since they can end early
```
hschawe, line 80: when generateResponseWithTools calls a custom tool, it returns the tool call instead of continuing generation, so if a custom tool is called before our hidden tools then no context is retrieved for the eval case. It's not too impactful on our metrics but just want to get y'alls thoughts

packages/chatbot-server-mongodb-public/src/conversations.eval.ts
```diff
@@ -35,12 +35,33 @@ async function conversationEval() {
     datasetName: "voyage-ai",
   });
   assert(voyageCases.length > 0);
+  const systemPromptCases = await getConversationsEvalCasesFromYaml(
+    fs.readFileSync(
+      path.resolve(basePath, "system_prompt_conversations.yml"),
+      "utf8"
+    )
+  );
+  const secretToolCases = await getConversationsEvalCasesFromYaml(
```
mongodben, line 44: i kno i said in the meeting the other day that we should call them 'secret tools' but immediately afterwards i thought that was a bad name. instead lets brand as 'internal tools'

packages/chatbot-server-mongodb-public/evalCases/custom_tool_conversations.yml
```diff
@@ -0,0 +1,182 @@
...
+  customTools:
+    - name: code_execution
+      description: Executes a given code snippet and returns the output. This tool should be called when debugging  or troubleshooting to get code execution results
+      parameters:
+        code:
+          type: string
+          description: The code snippet to execute.
+        language:
+          type: string
+          description: The programming language of the code snippet.
```
mongodben, line 41: nit: i'd reverse the order on these w/ `language` before `code` b/c the model will generate that first

packages/chatbot-server-mongodb-public/evalCases/custom_tool_conversations.yml
mongodben, line 1: i wonder if yaml is the right format for these. b.c the tool defs can be reused across cases if you're using JS/TS objects.

no action nec needed now/ever

packages/chatbot-server-mongodb-public/src/eval/ConversationEval.ts
```diff
@@ -185,6 +199,246 @@ const makeFactuality: ConversationEvalScorerConstructor =
...
+  ]
+}
+
+Do not return any preamble or explanations, return only a pure JSON string.
+`;
+
+const makePromptAdherence: ConversationEvalScorerConstructor = (
+  judgeModelConfig
+) => {
+  const judgeModel = wrapOpenAI(new AzureOpenAI(judgeModelConfig.azureOpenAi));
```
mongodben, line 227: pls use vercel ai sdk generateObject() function instead of this. trying to move all LLM calls to AI SDK. will make code a bit cleaner too

packages/chatbot-server-mongodb-public/src/eval/ConversationEval.ts
```diff
@@ -185,6 +199,246 @@ const makeFactuality: ConversationEvalScorerConstructor =
...
+        return messageScore;
+      })
+      .reduce((acc, score) => acc + score, 0) / args.output.messages.length;
+
+  return {
+    name: name,
+    score,
+    metadata: {
+      message: `Scored ${score} based on matching roles`,
+    },
```
mongodben, line 373: this metadata seems low value. maybe instead could return the total num matches and the total num messages:

```ts
{
  //...
  metadata: { numMatches: 3, totalMessages: 5}
}
```

packages/chatbot-server-mongodb-public/src/systemPrompt.ts
```diff
@@ -69,6 +69,7 @@ const searchContentToolNotes = [
 
 const fetchPageToolNotes = [
   "Fetch the entire page content for a given URL.",
+  `Do not assume that the user wants to use the ${FETCH_PAGE_TOOL_NAME} based on the URL in the Front Matter. The ${FETCH_PAGE_TOOL_NAME} should ONLY be used if the user implies you should look on the page or if the user explicitly provides a URL in their question.`,
```
mongodben, line 72:
```suggestion
  `Do not assume that the user wants to use the ${FETCH_PAGE_TOOL_NAME} based on URL present in the Front Matter. The ${FETCH_PAGE_TOOL_NAME} should ONLY be used if the user implies you should look on the page or if the user explicitly provides a URL in their question.`,
```

the messages from other clients won't nec have frontmatter, i think. will we need to update prompt elsewhere to take this into account?

packages/mongodb-rag-core/src/eval/getConversationEvalCasesFromYaml.ts
```diff
@@ -1,19 +1,52 @@
...
     .string()
     .optional()
+    .describe("A description of what the function does."),
+  parameters: z
+    .record(
+      z.string(),
+      z.object({
+        type: z.string().optional(),
+        description: z.string().optional(),
+      })
```
mongodben, line 20: i dont think this is correct tyling b/c it can be deeply nested, have additional props like min, max, etc. 

instead i think just z.any() is fine here

mongodben: unless you can find a zod type that represents all valid json schema (which might exist given popularity of both tools)....don't sweat this tho. z.any() should be fine

packages/chatbot-server-mongodb-public/src/conversations.eval.ts
hschawe: agreed!

mongodben (approved):
nice work!

PR #897, opened 2025-08-22
(EAI-1187) Instruct chatbot on handling fetch_page fallback

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1187

## Changes

- Let user know that we searched after failing to find info in an on-page reference (fetch_page fallback to search case)

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/systemPrompt.ts
+++ b/packages/chatbot-server-mongodb-public/src/systemPrompt.ts
@@ -18,6 +18,9 @@ export type MakeSystemPrompt = (
 export const llmDoesNotKnowMessage =
   "I'm sorry, I do not know how to answer that question. Please try to rephrase your query.";
 
+const fallbackDisclaimer =
+  "I didn't find anything on the page you provided, so I did a search of all MongoDB content instead.";
+
 const chatbotOverview = `You are expert MongoDB documentation chatbot.`;
 
 const personalityTraits = [
@@ -44,6 +47,7 @@ const technicalKnowledge = [
 
 const importantNotes = [
   `ALWAYS use either ${SEARCH_TOOL_NAME} or ${FETCH_PAGE_TOOL_NAME} after every user message. Zero exceptions!`,
+  `If you used the ${FETCH_PAGE_TOOL_NAME} tool and had to fall back to ${SEARCH_TOOL_NAME}, tell the user you didn't find anything on the page they provided, and that you searched all MongoDB content instead.`,
 ];
 
 const metadataNotes = [
@@ -93,6 +97,9 @@ ${makeMarkdownNumberedList(personalityTraits)}
 
 If you do not know the answer to the question, respond only with the following text:
 "${llmDoesNotKnowMessage}"
+
+When crafting your final answer to the user, follow these guidelines:
+${importantNote}
 
 Response format:
 ${makeMarkdownNumberedList(responseFormat)}


packages/chatbot-server-mongodb-public/src/systemPrompt.ts
```diff
@@ -94,6 +95,9 @@ ${makeMarkdownNumberedList(personalityTraits)}
 If you do not know the answer to the question, respond only with the following text:
 "${llmDoesNotKnowMessage}"
 
+When crafting your final answer to the user, follow these guidelines:
+${importantNote}
```
mongodben, line 99: i dont think we wanna add this here. currently we only include the important notes if there isn't a custom tool call. whereas, here it'd always be included. 

instead of adding this bullet as an 'importantNote' can you instead just add it as one of the bullets in the tool description of the `fetch_page` tool?

mongodben (changes requested):
small request for change

packages/chatbot-server-mongodb-public/src/systemPrompt.ts
hschawe: good point, I hadn't noticed that `importantNote` was used that way. however when I put this response instruction in the `fetchPageToolNotes`, it gets ignored, so I've made this its own instruction (separate from `importantNote`) and kept it where it is now

packages/chatbot-server-mongodb-public/src/systemPrompt.ts
```diff
@@ -76,6 +76,10 @@ const fetchPageToolNotes = [
   `If the ${FETCH_PAGE_TOOL_NAME} tool returns the string "${SEARCH_ALL_FALLBACK_TEXT}", you MUST immediately call the ${SEARCH_TOOL_NAME} tool.`,
 ];
 
+const finalAnswerContentNotes = [
+  `If the user provided a URL but you had to use the ${SEARCH_TOOL_NAME} tool to answer their query, you MUST tell the user you didn't find anything on the page provided and that you searched your knowledge base instead.`
```
mongodben, line 80:
```suggestion
  `If the user provided a URL but you had to use the ${SEARCH_TOOL_NAME} tool to answer their query, you MUST tell the user you didn't find anything on the page provided and that you searched your knowledge base instead.`
```

should you mention that this occurs only when the FETCH_PAGE_TOOL shows no results?

packages/chatbot-server-mongodb-public/src/systemPrompt.ts
```diff
@@ -138,28 +136,27 @@ ${makeMarkdownNumberedList(searchRequiresRephraseNotes)}
 
 <tools>
 
-<tool name="${SEARCH_TOOL_NAME}">
+You have access to the search_content and fetch_page tools.
```
mongodben, line 139:
```suggestion
You have access to the ${SEARCH_TOOL_NAME} and ${FETCH_PAGE_TOOL_NAME} tools.
```

mongodben (approved):
LGTM, great prompt engineering work here!

1 small suggestion

mongodben:
>   1. OpenAI's [GPT-4.1 prompting guide](https://cookbook.openai.com/examples/gpt4-1_prompting_guide#tool-calls:~:text=Developers%20should%20name%20tools%20clearly%20to%20indicate%20their%20purpose%20and%20add%20a%20clear%2C%20detailed%20description%20in%20the%20%22description%22%20field%20of%20the%20tool.) recommends to put all tool instructions in the tool descriptions, and I saw a notable increase across all metrics after doing this (for fetch_page and prompt adherence eval cases)

great, in the prompting guide and evals i trust.

interesting that this is the optimal pattern now. it actually used to be that the API had a hard cap on how long the description could be, and you were _supposed_ to put all this info in the system prompt. wonder when it changed.

>   2. There were still some issues with selecting the right tool & formatting the responses, so I put the "coordination" instructions in the system prompt, which improved prompt adherence  & the tool call metrics

sounds good

>   3. Repeating instructions greatly helps the LLM follow instructions when it comes to tool instructions. I tested out some search fallback changes (which did not make it into this PR) and the fetch_page assumption instructions (which did) and in both cases having an extra line of instruction really helped the model use the tools correctly

👍 

> * The LLM isn't always great at always giving the fallback response instruction in only the {fallback_to_search} case.  Sometimes when fetch_page has to truncate the page, the LLM will sometimes say it couldn't use the page, even if it didn't call the search_tool.

sorry i dont really understand what you mean here. can you explain a bit more?

and, is this an issue that we should address? if so, with what priority? should there be a follow up ticket?

packages/chatbot-server-mongodb-public/src/systemPrompt.ts
hschawe: whoops !

hschawe:
> > * The LLM isn't always great at always giving the fallback response instruction in only the {fallback_to_search} case.  Sometimes when fetch_page has to truncate the page, the LLM will sometimes say it couldn't use the page, even if it didn't call the search_tool.
> 
> sorry i dont really understand what you mean here. can you explain a bit more?
> 
> and, is this an issue that we should address? if so, with what priority? should there be a follow up ticket?

when `fetch_page` is used on a long page (>150,000 characters), the page is truncated and we do an on-page search for relevant content. the LLM knows that this search is happening, so sometimes - not always - it will give the fallback disclaimer, even though it didn't fall back to the `search_content` tool. 

i don't think this is *that* big of an issue especially since it doesn't affect how the LLM uses the tools, but it would be good to dedicate some time to this on a separate ticket

PR #900, opened 2025-08-26
(EAI-1279): dataset sampling args

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-1279

## Changes

- Add `--sampleSize` and `--sampleType` args to get subset of dataset
  - These are quite useful when iterating on prompts.
- Fix bug in `benchmarkCli.test.ts` where tests were failing w/ a different err type than expected (needed to pass a valid `--model` arg to all cases)

## Notes

-


hschawe (approved):
LGTM

PR #904, opened 2025-08-27
(EAI-1284): fix analyzer API bug

mongodben:
Jira: https://jira.mongodb.org/browse/EAI-1284

## Changes

- Fix bug where analyzer model wasn't working b/c of bad API version. Fix by using a different OpenAI client. Remove broken one
- Remove unused preprocessor client

## Notes

-

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/config.ts
+++ b/packages/chatbot-server-mongodb-public/src/config.ts
@@ -121,13 +121,6 @@ export const openAiClient = wrapOpenAI(
   })
 );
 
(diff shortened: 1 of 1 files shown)


packages/chatbot-server-mongodb-public/src/config.ts
```diff
@@ -220,13 +213,6 @@ export const loadPage = wrapTraced(pageStore.loadPage, {
   name: "loadPageFromStore",
 });
```
mongodben, line 215: this was unused

PR #910, opened 2025-09-02
(EAI-1150) Improve cleanliness in generateResponseWithTools.ts

hschawe:
Jira: https://jira.mongodb.org/browse/EAI-1150

## Changes

- handle tool call outside of generation loop
- pass the tools into the function as a list instead of individually

## Notes

- I decided to keep the tool result references handling in the generation loop. It's quick and the current code is generic for all tools.

### DIFF
--- a/packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.ts
+++ b/packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.ts
@@ -39,11 +39,8 @@ import {
   GenerateResponseParams,
 } from "mongodb-chatbot-server";
 import { formatUserMessageForGeneration } from "../processors/formatUserMessageForGeneration";
-import { MongoDbSearchToolArgs, SEARCH_TOOL_NAME } from "../tools/search";
-import {
-  FETCH_PAGE_TOOL_NAME,
-  MongoDbFetchPageToolArgs,
-} from "../tools/fetchPage";
+import { SEARCH_TOOL_NAME, SearchTool } from "../tools/search";
+import { FETCH_PAGE_TOOL_NAME, FetchPageTool } from "../tools/fetchPage";
 import { MakeSystemPrompt } from "../systemPrompt";
 import { logRequest } from "../utils";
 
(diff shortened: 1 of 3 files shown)


mongodben (changes requested):
this is def an improvement over the current state but i think it's likely that we can delete more code here. the current implementation always felt a bit hackish to me, and is more an artifact of when we were doing non-agentic RAG. i think w/ the current agentic tool calling behavior, we dont need to stuff all this in the user message

 see comment for more https://github.com/mongodb/chatbot/pull/910/files#r2319368961

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.ts
```diff
@@ -630,14 +613,6 @@ export function makeGenerateResponseWithTools({
...
         userMessage.rejectQuery = guardrailResult.rejected;
-        userMessage.metadata = {
-          ...userMessage.metadata,
-        };
-        userMessage.customData = {
-          ...userMessage.customData,
-          ...userMessageCustomData,
-          ...guardrailResult,
-        };
         if (streamingModeActive) {
```
mongodben, line 616: on line 639 i think we should still spread in the `guardrailResult` since that info isn't anywhere else

packages/chatbot-server-mongodb-public/src/tracing/extractTracingData.ts
```diff
@@ -162,3 +166,25 @@ export function getContextsFromMessages(
 
   return contexts;
 }
+
+export const getFirstToolMetadata = (
```
mongodben, line 170: why are we just going for the first tool?

hschawe: happy to bring back the multi-tool metadata extraction fxn but i see it as somewhat unnecessary. couple thoughts on why:
- i only see us using tool call metadata here as a sanity check, especially for the LLM-generated search query, etc. we have braintrust for more in-depth debugging
- while it's possible for multiple search_content tool calls to be returned by the LLM (based on the system prompt), i have never observed this happening
- pretty similar for fetch_page, it could return multiple tool calls but those cases should be obvious from the user's question. for analysis, the useful part is the search query
- and the search->fetch fallback almost always uses the same search query so we'd be storing that twice

packages/chatbot-server-mongodb-public/src/processors/generateResponseWithTools.ts
hschawe: the guardrail result is added to customData in `handleReturnGeneration`!

